PR #5, opened 2021-04-28
Run Github Actions on PR trigger

### LINKED ISSUE #4: Github Actions running twice on Pull Requests
This is likely due to the rule:

https://github.com/gnosis/safe-config-service/blob/227df740327663368a191d2aa2f49aa59f28d132/.github/workflows/test.yml#L3

It should run on each new commit once and only once.

fmrsabino:
Closes #4 

- Execute Github Action on PR event (instead of push+pr event which would trigger the action twice on pull requests)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,10 @@
 name: "CI"
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
 
 jobs:
   test:


Uxio0 (approved):

.github/workflows/test.yml
```diff
@@ -1,6 +1,10 @@
 name: "CI"
 
-on: [push, pull_request]
+on:
```
rmeissner, line 3: Wouldn't `push` have been alright to? 

I think the difference would be:
- `push` -> you need to be able to push to our repo
- `pull_request` -> anyone can create a PR to our repo

fmrsabino: Good point. So the way it is right now means that the pipeline will run on:
- `push` events to the `main` branch
- `pull_request` events (`opened`, `synchronize` and `reopened` [by default](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request))

This means that other branches (besides `main`) would not run the pipeline until a Pull Request is open. So I think we can either enable to all the branches and your suggestion for `push` would suffice or leave it as it is.

rmeissner: > all the branches
this would be all branches **in our repo**. As we normally don't create random branches and push them I don't think this would be an issue.

I do not have a strong opinion and I am fine either way. But I would keep that in mind as this repo will contain security critical secrets.

Uxio0: For what I tested on the services no one inside our organization will have access to the secrets, even if they can open a PR. Only way for people external to access the secrets would be allowing them as collaborators on the repo

fmrsabino: It seems that [`pull_request_target`](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target) has access to secrets even when triggered by forks (which was not what was applied here).

But maybe converting it to just `push` simplifies it.

Uxio0: Yes, but we were using `pull_request`, right?

fmrsabino: > Yes, but we were using `pull_request`, right?

Yes 👍

PR #6, opened 2021-04-28
Add initial default apps fixture

### LINKED ISSUE #2: Bootstrap database with default data for the /safe-apps endpoint
When the database is created and the service is running, we should offer the option/utility to bootstrap it with the default list of safe-apps.

Default database should contain the following data: https://github.com/gnosis/safe-apps-list/blob/development/public/gnosis-default.applist.json

fmrsabino:
Closes #2 

- Adds a default initial list for the `/api/v1/safe-apps/` endpoint. The list was taken from https://github.com/gnosis/safe-apps-list/blob/development/public/gnosis-default.applist.json
- The list can be loaded to the database via `python src/manage.py loaddata initial_default_apps.json`
  * It does not delete already existing data
  * It updates entries that match the url (primary key) – ie.: the value of this fixture overrides the current entry already set in the database (if any)
- Primary key is now the application URL (removed the auto-generated primary key)


Uxio0 (approved):
I think it's a good solution for a first version, but from here if we want to change anything that's hardcoded (maybe for people hosting this application or even ourselves in other environments) we need to use migrations or a custom command

PR #15, opened 2021-04-30
Improve admin display for Provider and SafeApp

### LINKED ISSUE #11: Improve admin page
- [x] Add custom Admin models with `list_display`, `display_filter` (for example, by network), `ordering` and `search_fields`: https://docs.djangoproject.com/en/3.2/ref/contrib/admin/
- [x] Use admin decorator https://docs.djangoproject.com/en/3.2/ref/contrib/admin/#the-register-decorator

fmrsabino:
Closes #11 

- Add `list_display`, `list_filter`, `search_fields` and `ordering` to the admin pages of `SafeApp` and `Provider`
- Add `NetworksFilter` to provider a filter over the array of networks for each app
  * Since each `SafeApp` contains a list of networks, this list needs to be flattened and converted to a set
- Use `@admin.register` decorator for `SafeApp` and `Provider`

### DIFF
--- a/src/safe_apps/admin.py
+++ b/src/safe_apps/admin.py
@@ -12,12 +12,10 @@ class NetworksFilter(admin.SimpleListFilter):
         # lookups requires a tuple to be returned – (value, verbose value)
         networks = [(network, network) for networks in values for network in networks]
         networks = sorted(set(networks))
-        print(networks)
         return networks
 
     def queryset(self, request, queryset):
-        value = self.value()
-        if value:
+        if value := self.value():
             queryset = queryset.filter(networks__contains=[value])
         return queryset
 
--- a/src/safe_apps/admin.py
+++ b/src/safe_apps/admin.py
@@ -2,5 +2,36 @@ from django.contrib import admin
 
 from .models import SafeApp, Provider
 
-models = [SafeApp, Provider]
-admin.site.register(models)
+
+class NetworksFilter(admin.SimpleListFilter):
+    title = 'Networks'
+    parameter_name = 'networks'
+
+    def lookups(self, request, model_admin):
+        values = SafeApp.objects.values_list('networks', flat=True)
+        # lookups requires a tuple to be returned – (value, verbose value)
+        networks = [(network, network) for networks in values for network in networks]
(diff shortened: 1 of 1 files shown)


src/safe_apps/admin.py
```diff
@@ -2,5 +2,36 @@
...
+class NetworksFilter(admin.SimpleListFilter):
+    title = 'Networks'
+    parameter_name = 'networks'
+
+    def lookups(self, request, model_admin):
+        values = SafeApp.objects.values_list('networks', flat=True)
+        # lookups requires a tuple to be returned – (value, verbose value)
+        networks = [(network, network) for networks in values for network in networks]
+        networks = sorted(set(networks))
+        print(networks)
```
Uxio0, line 15: Remove print

src/safe_apps/admin.py
```diff
@@ -2,5 +2,36 @@
...
+        return networks
+
+    def queryset(self, request, queryset):
+        value = self.value()
+        if value:
```
Uxio0, line 20: Since python 3.8 you can do `if value := self.value():` (just as a tip, but this way is ok and more compatible with previous versions)

src/safe_apps/admin.py
```diff
@@ -2,5 +2,36 @@
...
+            queryset = queryset.filter(networks__contains=[value])
+        return queryset
+
+
+@admin.register(SafeApp)
+class SafeAppAdmin(admin.ModelAdmin):
+    list_display = ('name', 'url', 'networks')
+    list_filter = (NetworksFilter,)
```
Uxio0, line 28: Setting here `networks` does not create the list_filter for you?

Uxio0 (changes requested):

src/safe_apps/admin.py
fmrsabino: The list was created but unfortunately only with the different sets:



By flattening the list with `NetworksFilter` we can have the following set:

Uxio0: Nice, thanks, just curious!

Uxio0 (approved):

PR #16, opened 2021-04-30
Use FactoryBoy for generating test data

### LINKED ISSUE #9: Use Factory Boy for testing models
Instead of https://github.com/gnosis/safe-config-service/blob/main/src/safe_apps/tests/test_views.py#L49 use https://factoryboy.readthedocs.io/en/stable/orms.html#django

fmrsabino:
Closes #9 

- Adds a factory for `Provider` and `SafeApp` so that test data can be generated more easily

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,4 @@
 -r requirements.txt
 pytest-django==4.2.0
 flake8==3.9.1
+factory-boy==3.2.0
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,4 @@
 -r requirements.txt
-pytest-django==4.2.0
-flake8==3.9.1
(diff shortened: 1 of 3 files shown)


requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
 -r requirements.txt
 pytest-django==4.2.0
 flake8==3.9.1
+factory-boy==3.2.0
```
Uxio0, line 4: Try to have the requirements files sorted, not important but easier to look at 😉

Uxio0 (approved):
Well done! I think it's easier to use factories, what do you think?

Uxio0 (approved):

fmrsabino:
> Well done! I think it's easier to use factories, what do you think?

Yes completely agree! 🙌

PR #23, opened 2021-05-04
Use Nginx templates to replace environment variables

### LINKED ISSUE #22: Set port numbers in docker compose via environment variables
To provide more flexibility when setting up the ports for a given environment, we should not use static ports in `docker-compose`. Instead those ports should be extracted to the `.env` file.

fmrsabino:
Closes #22 

- Extract port variables to the `.env` file
- Use Nginx templating system
  * This allows that one template of a nginx configuration with environment variables correctly generates a configuration with those environment variables set
  * The nginx image will look for those templates in the image location `/etc/nginx/templates`
  * The `NGINX_ENVSUBST_OUTPUT_DIR` is an environment variable used by nginx – the path set here is the path where the generated configs (with the environment variables replaced) will be stored

### DIFF
--- a/nginx/templates/nginx.conf.template
+++ b/nginx/templates/nginx.conf.template
@@ -25,7 +25,7 @@ http {
       # server unix:/run/gunicorn.sock fail_timeout=0;
 
       # for a TCP configuration
-      server web:8000 fail_timeout=0;
+      server web:${GUNICORN_BIND_PORT} fail_timeout=0;
       keepalive 32;
     }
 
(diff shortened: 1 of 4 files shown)


nginx/templates/nginx.conf.template
```diff
@@ -25,7 +25,7 @@ http {
       # server unix:/run/gunicorn.sock fail_timeout=0;
 
       # for a TCP configuration
-      server web:8000 fail_timeout=0;
+      server web:${GUNICORN_BIND_PORT} fail_timeout=0;
```
fmrsabino, line 28: The location and name of the file is important for the nginx image.

Location – used as a volume in docker-compose
Name – nginx will look for `.template` files and generate a file with the suffix removed. In this case `nginx.conf`

Uxio0 (approved):
I didn't know this was possible with nginx! Nice!

fmrsabino:
> I didn't know this was possible with nginx! Nice!

I think it was added recently in image version 1.19: https://hub.docker.com/_/nginx (section `Using environment variables in nginx configuration (new in 1.19)`)

Found out when was extracting more environment variables to make it easier to deploy with a specific configuration 😅

Uxio0:
That's a really good find, thanks!

PR #27, opened 2021-05-05
Add Logging middleware

### LINKED ISSUE #26: Integrate logging middleware
The service should integrate with a Logging Middleware in order to follow the standard logging structure that https://github.com/gnosis/safe-transaction-service follows.

See Logging Middleware here: https://github.com/gnosis/safe-transaction-service/blob/ed0720ce238153e678d0e6df5d48707c7be99fdd/config/settings/base.py#L91

fmrsabino:
Closes #26 

- Add a Logging middleware – this middleware logs endpoint request calls to `INFO`
- Add Django logging settings
  * A `root` logger that applies the configuration to loggers which are not declared in this dictionary. Default logging level is `INFO` (can be changed by setting `ROOT_LOG_LEVEL`)
  * A `LoggingMiddleware` that logs view requests to `INFO`. It currently logs the HTTP method used, the route and paht, the time that it took to process the request and the status of the response.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -14,6 +14,10 @@ PYTHONDONTWRITEBYTECODE=true
 # This should never be set to true in production but it should be enabled in dev.
 DEBUG=false
 
+# Root log level (default is INFO)
+# Possible values are DEBUG | INFO | WARNING | ERROR | CRITICAL
+ROOT_LOG_LEVEL=INFO
+
 # A comma separated list of allowed hosts. In production this should be your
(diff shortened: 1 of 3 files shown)


fmrsabino:
> Awesome. Do you think https://pypi.org/project/python-environ/ could be useful for you to load the environment variables?

Looks interesting! I think right now the configuration is still quite simple to add this dependency. So I'd leave it for now but can be considered in the future :)

PR #29, opened 2021-05-06
Propagate subcommand failure return code

fmrsabino:
- Propagate return code of each sub-command to the pipeline (by default it was taking the last one executed, in this case the tests)
- The `-eo pipefail` set the pipeline to exit earlier if if a return code is `>0` (ie.: not successful)

### DIFF
--- a/run
+++ b/run
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -eo pipefail
+
 DC="${DC:-exec}"
 
 # If we're running in CI we need to disable TTY allocation for docker-compose


fmrsabino:
This was an overlook on my side. With the git hooks the formatting was done correctly and would never fail. The pipeline always failed if the tests failed but the result of previous steps would be ignored.

Uxio0 (approved):
I missed that too

fmrsabino:
> I missed that too

Initially the command was mostly running the tests which would work. The error results of the commands that were added were shown in the logs but not propagated to the pipeline 😬

PR #30, opened 2021-05-06
Cache GET /safe-apps/ endpoint

### LINKED ISSUE #28: Cache GET /safe-apps/ endpoint
The endpoint `GET /safe-apps` returns a list of safe-apps that can be cached (eg.: 1h) due to the nature of the endpoint (frequency of updating the app list is low). Updating the list should invalidate the cache.

fmrsabino:
Closes #28 

- Cache `safe-apps` endpoint for 1h
- Update tests that do not required caching to use `DummyCache`
- Add test to test caching

### Some notes on key invalidation:

- It seems that adding view caching is relatively straightforward with `cache_page` but removing the cache entry when new data is added/updated/deleted seems to be complex
- The complexity comes from how Django generates the keys for cache entries they are based on the request that was made, the environment where the request was made, and header properties – https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/django/utils/cache.py#L347-L366
- Correctly recreating the key is possible as the information is deterministic but it'd be a custom solution on our side that might not work on later versions. Given this I decided to not invalidate those cache entries on these events.

### DIFF
--- a/src/safe_apps/views.py
+++ b/src/safe_apps/views.py
@@ -9,7 +9,7 @@ from .serializers import SafeAppsResponseSerializer
 class SafeAppsListView(ListAPIView):
     serializer_class = SafeAppsResponseSerializer
 
-    @method_decorator(cache_page(60 * 60 * 1))  # Cache 1 hour
+    @method_decorator(cache_page(60 * 10))  # Cache 10 minutes
     def get(self, request, *args, **kwargs):
         return super().get(self, request, *args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


src/safe_apps/views.py
```diff
@@ -7,6 +9,10 @@
 class SafeAppsListView(ListAPIView):
     serializer_class = SafeAppsResponseSerializer
 
+    @method_decorator(cache_page(60 * 60 * 1))  # Cache 1 hour
```
fmrsabino, line 12: For now it's set for 1h. But up for some suggestions. Can also be extracted to an environment variable if we see fit.

Uxio0 (approved):

Uxio0 (approved):

PR #32, opened 2021-05-07
Use different cache for view. Invalidate on update

### LINKED ISSUE #31: Invalidate cache on SafeApp model update
The cache should be invalidated on model updates (create, update, delete). This can be accomplished by storing the cache entry in a separate cache (due to the complexity of retrieving the correct cache key) and clearing that cache.

fmrsabino:
Closes #31 

- Add `safe-apps` cache (which is a `LocMemCache`)
- Add signal which is triggered on `SafeApp` or `Provider` updates
- Test now checks if the response header `Cache-Control` returns the correct value

### DIFF
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -53,6 +53,15 @@ MIDDLEWARE = [
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
 ]
 
+CACHES = {
+    "default": {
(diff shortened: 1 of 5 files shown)


Uxio0 (dismissed):

Uxio0:
Looks like a really good solution

Uxio0 (approved):

Uxio0 (approved):

PR #36, opened 2021-05-07
Refactor Github Actions

### LINKED ISSUE #35: Improve Github Actions workflow
- Do not use Docker for running tests and linter
- Cache dependencies
- Enable parallel workflows whenever possible

fmrsabino:
Closes #35 

- Use Github Actions to run Linter and Tests (ie.: avoiding running these inside the docker image)
- Runs 4 jobs in parallel:
  * `flake8`
  * `isort`
  * `black`
  * `django-check` – runs postgres image, executes migrations, django system check and runs integration tests
- Removes functions from bash script that were being used to run the tasks listed above insider the container. They can still be run locally using `./run cmd <task>` (if running the docker images locally)

Note: A good outcome of this refactor is that these steps were taking 2minutes+ and now it takes just under a minute. Example run: https://github.com/gnosis/safe-config-service/actions/runs/821080982


fmrsabino:
> Nice, I didn't know about liskin/gh-problem-matcher-wrap

Yeah 👍 Seems that the Django project is also using it :)

PR #38, opened 2021-05-10
Add Docker deploy step for main branch

fmrsabino:
- Deploys production docker image to Docker Hub when the `main` branch is updated
- Cache docker layers with `cache-from` and `cache-to`
- Ignore files which are not required for the docker web image

### DIFF
--- a/.github/workflows/publish-staging.yml
+++ b/.github/workflows/publish-staging.yml
@@ -32,7 +32,7 @@ jobs:
           push: true
           # Target production image from multi-stage build (check Dockerfile)
           target: python-base
-          tags: gnosispm/safe-config-service:latest
+          tags: gnosispm/safe-config-service:staging
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache-new
       - # Temp fix
--- a/.github/workflows/publish-staging.yml
+++ b/.github/workflows/publish-staging.yml
@@ -16,8 +16,8 @@ jobs:
       - name: Login to DockerHub
         uses: docker/login-action@v1
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-staging.yml
```diff
@@ -0,0 +1,46 @@
...
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Build and push
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: true
+          # Target production image from multi-stage build (check Dockerfile)
+          target: python-base
+          tags: gnosispm/safe-config-service:latest
```
fmrsabino, line 35: @Uxio0 images built on the `main` branch will be published to Docker Hub with the `latest` tag. Let me know if we should change it to something else

fmrsabino:
And `docker/setup-buildx-action` is quite amazing. Caching the layers of the image reduced a 2m14s build down to 44s 🚀

.github/workflows/publish-staging.yml
Uxio0: That's reasonable

Uxio0:
> And `docker/setup-buildx-action` is quite amazing. Caching the layers of the image reduced a 2m14s build down to 44s 🚀

I need to try that!

Uxio0 (approved):

.github/workflows/publish-staging.yml
fmrsabino: Actually just synced with @giacomolicari and this was changed to `staging` to be in line with the other services 👍

Uxio0: Makes sense. But they are just tags, you can even push both

PR #39, opened 2021-05-10
Continue pipeline if coverage upload fails

fmrsabino:
Adding the `continue-on-error` to avoid an external service being down and blocking the PR


Uxio0 (approved):
I did the same, as coveralls was failing

fmrsabino:
> I did the same, as coveralls was failing

My inspiration came from there 😄

PR #40, opened 2021-05-11
Add missing steps to production image

fmrsabino:
- Previously the Dockerfile was split into 2 stages:
  * `python-base` – base dependencies to build the project + installation of production dependencies
  * `development` – base + development dependencies

One issue here is that `python-base` didn't copy the application to the image and this was the stage that was being published to Docker Hub. To fix that issue and to make each stage more explicit these are the new stages:
  * `production` – base dependencies to build the project + copy of the application + production dependencies
  * `development` – production + installation of development dependencies

### DIFF
--- a/.github/workflows/publish-staging.yml
+++ b/.github/workflows/publish-staging.yml
@@ -31,7 +31,7 @@ jobs:
           context: .
           push: true
           # Target production image from multi-stage build (check Dockerfile)
-          target: python-base
+          target: production
           tags: gnosispm/safe-config-service:staging
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache-new
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,6 @@ services:
     tty: true
     env_file:
       - .env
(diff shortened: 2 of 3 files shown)


.github/workflows/publish-staging.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           context: .
           push: true
           # Target production image from multi-stage build (check Dockerfile)
-          target: python-base
+          target: production
```
fmrsabino, line 34: New target applied to the CI

docker-compose.yml
```diff
@@ -26,7 +26,6 @@ services:
     tty: true
     env_file:
       - .env
-    command: gunicorn -c python:config.gunicorn config.wsgi -b 0.0.0.0:${GUNICORN_BIND_PORT}
-    working_dir: /app/src
+    command: gunicorn -c /app/src/config/gunicorn.py config.wsgi -b 0.0.0.0:${GUNICORN_BIND_PORT} --chdir /app/src/
```
fmrsabino, line 29: Specified the full path for the configuration to avoid issues regarding the default entry point.

`--chdir` changes the base directory where it will find `config.wsgi`

Uxio0 (approved):

PR #44, opened 2021-05-12
Add entrypoint script

### LINKED ISSUE #43: Add ENTRYPOINT to the web app image
In order to make it easier to deploy and run the image, it should include the default entrypoint to the application

fmrsabino:
Closes #43 

- Adds a Docker `ENTRYPOINT` to the production image (also applies to the dev image).
- The entrypoint performs the following:
  * Executes migrations
  ~* Loads initial fixture data using `manage.py loaddata` (this does not override current data)~
  * Launches app with `Gunicorn`

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,0 +1,12 @@
+#!/bin/ash
+
+set -euo pipefail
+
+echo "==> $(date +%H:%M:%S) ==> Migrating Django models..."
+python src/manage.py migrate --noinput
+
+echo "==> $(date +%H:%M:%S) ==> Applying fixtures..."
+python src/manage.py loaddata initial_default_apps.json
+
+echo "==> $(date +%H:%M:%S) ==> Running Gunicorn..."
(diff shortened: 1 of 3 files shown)


docker-entrypoint.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/ash
+
+set -euo pipefail
+
+echo "==> $(date +%H:%M:%S) ==> Migrating Django models..."
+python src/manage.py migrate --noinput
+
+echo "==> $(date +%H:%M:%S) ==> Applying fixtures..."
+python src/manage.py loaddata initial_default_apps.json
```
Uxio0, line 9: Will this be removed in next versions?

Uxio0 (approved):

docker-entrypoint.sh
fmrsabino: Do you think it's better for us to load the fixture independently of the deployment? I can remove it on the context of this PR

fmrsabino: Removed here 46cc4dc

Uxio0: Yes, I think so. If not we can find surprises if we forget about it

Uxio0 (approved):

PR #58, opened 2021-05-14
Set default REST renderer

### LINKED ISSUE #57: Change default renderer to JSONRenderer
- The renderer in production should show the json payload without a browsable API

fmrsabino:
Closes #57 

- Sets default REST renderer to `rest_framework.renderers.JSONRenderer`

### DIFF
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -31,6 +31,17 @@ ALLOWED_HOSTS = list(map(str.strip, allowed_hosts.split(",")))
 
 # Application definition
 
+
+default_renderer_classes = os.getenv(
+    "REST_DEFAULT_RENDERER_CLASSES", "rest_framework.renderers.JSONRenderer"
+)
+REST_FRAMEWORK = {
+    # https://www.django-rest-framework.org/api-guide/renderers/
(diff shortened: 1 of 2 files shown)


src/config/settings.py
```diff
@@ -31,6 +31,17 @@
...
 # Application definition
 
+
+default_renderer_classes = os.getenv(
+    "REST_DEFAULT_RENDERER_CLASSES", "rest_framework.renderers.JSONRenderer"
+)
+REST_FRAMEWORK = {
+    # https://www.django-rest-framework.org/api-guide/renderers/
+    "DEFAULT_RENDERER_CLASSES": list(
+        map(str.strip, default_renderer_classes.split(","))
```
Uxio0, line 41: Please, instead of `map` use list comprehension `[default_renderer_class.strip() for default_renderer_class in default_renderer_classes.split(",")]` https://blog.finxter.com/which-is-faster-list-comprehension-or-map-function-in-python/

Uxio0 (changes requested):

src/config/settings.py
fmrsabino: A good catch 🙌

Uxio0 (approved):

PR #60, opened 2021-05-14
Use Unix socket in addition to TCP to bind application

### LINKED ISSUE #59: Use Unix socket instead of TCP to bind application
In order to avoid exposing service names when deploying the application we should use a socket to bind the connection between Nginx and the Gunicorn.

fmrsabino:
Closes #59 

- Use Unix socket to bind the application
- This avoids Nginx requiring to know the name registered of the container/service which is running the Gunicorn app

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -6,4 +6,4 @@ echo "==> $(date +%H:%M:%S) ==> Migrating Django models..."
 python src/manage.py migrate --noinput
 
 echo "==> $(date +%H:%M:%S) ==> Running Gunicorn..."
-exec gunicorn -c /app/src/config/gunicorn.py config.wsgi -b ${GUNICORN_BIND_SOCKET} --chdir /app/src/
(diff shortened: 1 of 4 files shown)


docker-entrypoint.sh
```diff
@@ -6,4 +6,4 @@ echo "==> $(date +%H:%M:%S) ==> Migrating Django models..."
 python src/manage.py migrate --noinput
 
 echo "==> $(date +%H:%M:%S) ==> Running Gunicorn..."
-exec gunicorn -c /app/src/config/gunicorn.py config.wsgi -b 0.0.0.0:${GUNICORN_BIND_PORT} --chdir /app/src/
+exec gunicorn -c /app/src/config/gunicorn.py config.wsgi -b ${GUNICORN_BIND_SOCKET} --chdir /app/src/
```
Uxio0, line 9: Do not remove the http entry, you can add both. It's good to have that so we can configure Kubernetes to check the pod without nginx

Uxio0 (changes requested):

Uxio0 (approved):

Uxio0 (approved):

PR #63, opened 2021-05-17
Fix lazy evaluation of ALLOWED_HOSTS and DEFAULT_RENDERER_CLASSES

### LINKED ISSUE #62: Fix lazy evaluation of ALLOWED_HOSTS and DEFAULT_RENDERER_CLASSES
`ALLOWED_HOSTS` and `DEFAULT_RENDERER_CLASSES` are both using list comprehensions which are not evaluated lazily. Instead they are evaluated when the image is built and those environment variables do not exist.

fmrsabino:
Closes #62 

ALLOWED_HOSTS and DEFAULT_RENDERER_CLASSES are both using list comprehensions which are not evaluated lazily. Instead they are evaluated when the image is built and those environment variables do not exist.

They were changed recently in the context of this PR: https://github.com/gnosis/safe-config-service/pull/58


fmrsabino:
This change is actually not required. List comprehension evaluates as we expect so no changes are required here

PR #65, opened 2021-05-17
Format JSON response keys to camel case

### LINKED ISSUE #64: Format JSON keys in camel case
Endpoints consumed by the clients should have the JSON keys in camel case. By having them camel case, it follows the formatting that we have in other services.

fmrsabino:
Closes #64 

- JSON keys in response payloads are now in camel case
- Add `djangorestframework-camel-case` – this renderer class is responsible for converting the keys to camel case (it is now the default renderer)

### Open Questions:

- Would this dependency be ok or shall we go with an internal solution? I saw already this dependency being used here so I guess we can continue using it: https://github.com/gnosis/safe-transaction-service/blob/cda755631cadbe03f2cc5e792d80f721212556c8/requirements.txt#L18

### DIFF
--- a/src/safe_apps/tests/test_views.py
+++ b/src/safe_apps/tests/test_views.py
@@ -14,6 +14,28 @@ class EmptySafeAppsListViewTests(APITestCase):
         self.assertEqual(response.data, [])
 
 
+class JsonPayloadFormatViewTests(APITestCase):
+    def test_json_payload_format(self):
+        safe_app = SafeAppFactory.create()
+
+        json_response = [
+            {
+                "url": safe_app.url,
+                "name": safe_app.name,
+                "iconUrl": safe_app.icon_url,
+                "description": safe_app.description,
+                "networks": safe_app.networks,
+                "provider": None,
+            }
+        ]
+        url = reverse("v1:safe-apps")
+
+        response = self.client.get(path=url, data=None, format="json")
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), json_response)
+
+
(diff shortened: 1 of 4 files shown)


Uxio0:
> * Would this dependency be ok or shall we go with an internal solution? I saw already this dependency being used here so I guess we can continue using it: https://github.com/gnosis/safe-transaction-service/blob/cda755631cadbe03f2cc5e792d80f721212556c8/requirements.txt#L18

I think it's ok. I've been using it for 3 years and no problems so far

src/safe_apps/tests/test_views.py
```diff
@@ -198,7 +198,7 @@ def test_should_cache_response(self):
             {
                 "url": safe_app_1.url,
                 "name": safe_app_1.name,
-                "icon_url": safe_app_1.icon_url,
+                "iconUrl": safe_app_1.icon_url,
                 "description": safe_app_1.description,
                 "networks": safe_app_1.networks,
                 "provider": None,
```
Uxio0, line 204: Just for your reference, if you want to test stuff without camel case you can do `response.data` and you have the data without the camel case renderer applied.

I would also prefer if you use `response.json()` instead of `json.loads(response.content)` 😉

Uxio0 (dismissed):

src/safe_apps/tests/test_views.py
fmrsabino: Updated 👍 Thanks for the tip!

Uxio0 (dismissed):

fmrsabino:
@Uxio0 I removed the environment variable and added a new test

Uxio0 (approved):

PR #66, opened 2021-05-17
Add CODEOWNERS

fmrsabino:
- Add `CODEOWNERS` to the project – see more here: https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
- Assign @fmrsabino and @Uxio0 as a code owners for the project

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,3 @@
 # These owners will be the default owners for everything in
 # the repo. Unless a later match takes precedence,
-* @fmrsabino
+* @fmrsabino @Uxio0
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,3 @@
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+* @fmrsabino


rmeissner:
I am not sure if I would do this. Having a dingle codeowner will block this when you are not around. Is there a specific reason why you think this is required?

fmrsabino:
> I am not sure if I would do this. Having a dingle codeowner will block this when you are not around. Is there a specific reason why you think this is required?

The Codeowner in this case just puts me automatically in reviews. It doesn't need my approval (even though that is a setting that can enabled) :)

It's more for visibility on my side.

image

rmeissner:
Shouldn't uxio also be a codeowner then? Also couldn't we do this via PR templates too (and even more)?

rmeissner:
I would put into the PR description what you want to archive with the PR (not only what code change have been made). As you do not have a linked issue the motivation is not directly clear. We should make sure to add this to the PR description, this will make it easier to review PRsnwithout a related issue ;)

fmrsabino:
> Shouldn't uxio also be a codeowner then? Also couldn't we do this via PR templates too (and even more)?

Yes I can also include Uxio. I'm not sure if I can request reviews via PR templates. I can probably cc people in the message but it would not be the same thing. We can potentially use a team as a reviewer if we see fit (`@safe-services`) as a reviewer for example.

In a larger project different sections might have different owners so that's why here the only rule right now would be: ` for all files request review from X`. I've created this more for visibility because PRs can be created without reviewers right now so this would be done automatically (except for draft PRs). But the scope is currently small so I also understand if it doesn't make sense right now

> I would put into the PR description what you want to archive with the PR (not only what code change have been made). As you do not have a linked issue the motivation is not directly clear. We should make sure to add this to the PR description, this will make it easier to review PRsnwithout a related issue ;)

Yes we should have a PR template for repo. Will create one 👍

Uxio0:
FYI: https://github.com/gnosis/safe-client-gateway/blob/main/.github/pull_request_template.md

Uxio0 (dismissed):

Uxio0 (approved):

PR #68, opened 2021-05-19
Add /check endpoint

### LINKED ISSUE #67: Add /check endpoint
This endpoint should be used to check if the server is alive and ready to receive requests

fmrsabino:
Closes #67 

- Adds `/check` endpoint that can be used to check the readiness for the service to receive requests

### DIFF
--- a/src/config/urls.py
+++ b/src/config/urls.py
@@ -1,9 +1,9 @@
 from django.contrib import admin
 from django.http import HttpResponse
-from django.urls import include, path, re_path
+from django.urls import include, path
 
 urlpatterns = [
     path("api/v1/", include("safe_apps.urls", namespace="v1")),
     path("admin/", admin.site.urls),
-    re_path(r"^check/", lambda request: HttpResponse("Ok"), name="check"),
+    path("check/", lambda request: HttpResponse("Ok"), name="check"),
 ]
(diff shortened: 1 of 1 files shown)


src/config/urls.py
```diff
@@ -1,7 +1,9 @@
 from django.contrib import admin
-from django.urls import include, path
+from django.http import HttpResponse
+from django.urls import include, path, re_path
 
 urlpatterns = [
     path("api/v1/", include("safe_apps.urls", namespace="v1")),
     path("admin/", admin.site.urls),
+    re_path(r"^check/", lambda request: HttpResponse("Ok"), name="check"),
```
Uxio0, line 8: Better use path instad of `re_path`, I should change it too :)

Uxio0 (changes requested):

src/config/urls.py
Uxio0: Well, I use `re_path` everywhere, so I might leave it that way as it's not deprecated

fmrsabino: Yeah I started with `path` but was wondering if I should change it to `re_path` as you have 🤔

fmrsabino: Can `path` have issues with ending the endpoint with `/` or without it? I've tested `check/` and `check` and both worked fine with `path`

Uxio0: There's an auto redirect, but stick with `check/`

Uxio0 (approved):

PR #80, opened 2021-05-27
Add CORS headers

### LINKED ISSUE #79: Missing CORS headers
Since CORS headers are missing, resources from this application cannot be accessed by other domains

fmrsabino:
Closes #79 

- Adds CORS headers
- Allow any origin to request the resources under `^/api/.*$`

### DIFF
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -39,6 +39,7 @@ REST_FRAMEWORK = {
 }
 
 INSTALLED_APPS = [
+    "corsheaders",
     "safe_apps.apps.AppsConfig",
     "django.contrib.admin",
     "django.contrib.auth",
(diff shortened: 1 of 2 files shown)


src/config/settings.py
```diff
@@ -182,3 +184,6 @@
 SWAGGER_SETTINGS = {
     "DEFAULT_INFO": "config.swagger_info.SAFE_CONFIG_SERVICE_SWAGGER_INFO"
 }
+
+CORS_ALLOW_ALL_ORIGINS = True
```
fmrsabino, line 188: @Uxio0 Thoughts on this one? This will set the header `Access-Control-Allow-Origin: *`

rmeissner: Same is done in the other services

See https://github.com/gnosis/safe-transaction-service/blob/master/config/settings/base.py#L161 and https://github.com/gnosis/safe-client-gateway/blob/main/src/utils/cors.rs#L21

rmeissner (approved):

src/config/settings.py
Uxio0: As @rmeissner said, we are using it as it's required if you don't want to be whitelisting client by client

PR #88, opened 2021-06-01
Add /api/v1/chains endpoint

### LINKED ISSUE #85: Add core chain information endpoint
Add chain information to the model and admin. The endpoint will be available under `/api/v1/chains`.
Each chain should have the following data:

```javascript
{
   "chainId": <stringified-int>, // stringified-int due to int size constraints
   "chainName": <string>, // optional
   "nativeCurrency": {
       "name": <string>, // optional
       "symbol": <string>,
       "decimals": <uint>,
   },
   "rpcUrl": <url-string>,
   "blockExplorerUrl": <url-string>
}
```

Optional fields are marked with `// optional`

⚠️ Note: this does not include parameters regarding client theme and `gnosis/safe-client-gateway`. The data represented here is only related with "core" chain data.

fmrsabino:
Closes #85 

- Add a new endpoint: `/api/v1/chains`
- This endpoint returns data about L2 chains that we support
- Swagger generation adjustments for the `/chains` endpoint

Json payload example:
```javascript
[
  {
    "chainId": "string",
    "chainName": "string",
    "rpcUrl": "string",
    "blockExplorerUrl": "string", // optional
    "nativeCurrency": {
      "name": "string",
      "symbol": "string",
      "decimals": 18
    }
  }
]
```

### Next steps:

This PR adds the model, views, serializers and default admin registration. For the next steps the following is going to be added:

- Adding pagination to the endpoint
- Improvements to the admin interface (sorting, filtering)
- Get a single chain by `id`


src/chains/views.py
```diff
@@ -0,0 +1,17 @@
...
+
+class ChainsListView(ListAPIView):
+    serializer_class = ChainSerializer
+
+    @swagger_auto_schema()
+    def get(self, request, *args, **kwargs):
+        return super().get(self, request, *args, **kwargs)
+
+    def get_queryset(self):
+        queryset = Chain.objects.all()
```
rmeissner, line 16: Should we inline this?

src/config/urls.py
```diff
@@ -11,7 +11,8 @@
 )
 
 urlpatterns = [
-    path("api/v1/", include("safe_apps.urls", namespace="v1")),
+    path("api/v1/", include("safe_apps.urls", namespace="safe-apps")),
```
rmeissner, line 14: what does the `namespace` do?

src/safe_apps/tests/test_views.py
```diff
@@ -6,7 +6,7 @@
 
 class EmptySafeAppsListViewTests(APITestCase):
     def test_empty_set(self):
-        url = reverse("v1:safe-apps")
+        url = reverse("safe-apps:safe-apps")
```
rmeissner, line 9: wouldn't something like `safe-apps:list_v1` be nicer?

rmeissner (approved):

src/config/urls.py
```diff
@@ -11,7 +11,8 @@
 )
 
 urlpatterns = [
-    path("api/v1/", include("safe_apps.urls", namespace="v1")),
```
fmrsabino, line 14: I've changed the namespace here. Since we are going to have more applications under `v1` I've split that namespace in two: `safe-apps` and `chains`.

By having:
```python
path("api/v1/", include("chains.urls", namespace="v1")),
```

We get the following warning:
```shell
(urls.W005) URL namespace 'v1' isn't unique. You may not be able to reverse all URLs in this namespace
```

src/config/urls.py
fmrsabino: This is mostly useful for the `reverse()` resolution of the url. So even if different included apps have the same pattern in the url (imagine something like `/index`) we can distinguish them by using `reverse()` with the `namespace`. Reverse is also to avoid using full paths – if the path changes, the url is correctly resolved to the view.

I've tried having everything under the `v1` namespace (left a comment about it here: https://github.com/gnosis/safe-config-service/pull/88#discussion_r643097398) but probably doesn't make too much sense to do it that way.

src/safe_apps/tests/test_views.py
fmrsabino: Yes I agree! 👍

fmrsabino: I'm going to do this change for the `chains:chains` and revert the changes on the `safe-apps` to reduce the scope of the PR. It should be only about the `chains` endpoint

src/config/urls.py
fmrsabino: I'm going to tackle the `safe-apps` namespace change in a different PR. For now I reverted that change here: 940bc16

fmrsabino: Created issue here: #89

src/safe_apps/tests/test_views.py
fmrsabino: Reverted here: 940bc16

Uxio0: You can go with url path versioning https://www.django-rest-framework.org/api-guide/versioning/#urlpathversioning

fmrsabino: > You can go with url path versioning https://www.django-rest-framework.org/api-guide/versioning/#urlpathversioning

But what about the root `urls.py`? What value would the two paths here have for the `namespace`?

```python
path("api/v1/", include("chains.urls", namespace="v1")),
```

fmrsabino: I can imagine something like:

```python
path("api", include("safe_apps.urls", namespace="safe-apps")),
path("api", include("chains.urls", namespace="chains")),
```

and inside each app do the versioning. Was this your suggestion @Uxio0 ?

Uxio0: If you use url inline you can do:

```python
urlpatterns = [
    re_path(
        r'^(?P<version>(v1|v2))/$',
        include("safe_apps.urls", namespace="safe-apps"),
    ),
    re_path(
        r'^(?P<version>(v1|v2))/$',
        include("chains.urls", namespace="chains"),
    )
]
```

Or you can do. I might prefer this version, looks cleaner:
```python
path("api/v1/", include("safe_apps.urls", namespace="v1:safe-apps")),
path("api/v1/", include("chains.urls", namespace="v1:chains")),
```

For the relay service I couldn't make it work and that's why I have everything on the same `urls` file (not recommended at all 🙈 )

fmrsabino: > The other option, I might prefer this version, looks cleaner:
> 
> ```python
> path("api/v1/", include("safe_apps.urls", namespace="v1:safe-apps")),
> path("api/v1/", include("chains.urls", namespace="v1:chains")),
> ```

If we go with this, and if eventually a new version gets introduced for example for `safe-apps` then it would be:

```python
path("api/v1/", include("safe_apps.urls", namespace="v1:safe-apps")),
path("api/v2/", include("safe_apps.urls", namespace="v2:safe-apps")),
path("api/v1/", include("chains.urls", namespace="v1:chains")),
```

right? 🤔

fmrsabino: I went with the first option that you suggested here @Uxio0  https://github.com/gnosis/safe-config-service/pull/88#discussion_r643234361

Since we don't have any other version besides `v1` I left it without the `re_path`. If a new version is added I will update the paths.

Uxio0: No, you would go with `path("api/v2/", include("safe_apps.urls_v2", namespace="v2:safe-apps")),` 😉

Uxio0 (approved):

src/safe_apps/tests/test_views.py
fmrsabino: Ah I see. Yeah that works 👍

PR #92, opened 2021-06-01
Add pagination to /chains endpoint

### LINKED ISSUE #91: Add pagination to the `chains/` endpoint
Add pagination support to `api/v1/chains`

fmrsabino:
Closes #91 

- Adds pagination to `api/v1/chains`
- Default pagination size is `10`, max pagination size is `10`

Pagination payload:

```javascript
{
  "count": 1,
  "next": null,
  "previous": null,
  "results": [
    {
      "chainId": "123123",
      "chainName": "Test",
      "rpcUrl": "http://example.com",
      "blockExplorerUrl": "http://example.com",
      "nativeCurrency": {
        "name": "Ether",
        "symbol": "ETH",
        "decimals": 18
      }
    }
  ]
}
```

### DIFF
--- a/src/chains/tests/test_views.py
+++ b/src/chains/tests/test_views.py
@@ -7,32 +7,93 @@ from .factories import ChainFactory
 class EmptyChainsListViewTests(APITestCase):
     def test_empty_chains(self):
         url = reverse("chains:list")
+        json_response = {"count": 0, "next": None, "previous": None, "results": []}
 
         response = self.client.get(path=url, data=None, format="json")
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.data, [])
+        self.assertEqual(response.data, json_response)
 
 
 class ChainJsonPayloadFormatViewTests(APITestCase):
     def test_json_payload_format(self):
         chain = ChainFactory.create()
-        json_response = [
-            {
-                "chainId": chain.id,
-                "chainName": chain.name,
-                "rpcUrl": chain.rpc_url,
-                "blockExplorerUrl": chain.block_explorer_url,
-                "nativeCurrency": {
-                    "name": chain.currency_name,
-                    "symbol": chain.currency_symbol,
-                    "decimals": chain.currency_decimals,
-                },
-            }
-        ]
+        json_response = {
+            "count": 1,
+            "next": None,
+            "previous": None,
+            "results": [
+                {
+                    "chainId": chain.id,
+                    "chainName": chain.name,
+                    "rpcUrl": chain.rpc_url,
+                    "blockExplorerUrl": chain.block_explorer_url,
+                    "nativeCurrency": {
+                        "name": chain.currency_name,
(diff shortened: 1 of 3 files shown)


src/chains/models.py
```diff
@@ -0,0 +1,14 @@
+from django.db import models
+
+
+class Chain(models.Model):
+    id = models.CharField(primary_key=True, max_length=255)
```
Uxio0, line 5: Why a CharField?

src/chains/views.py
```diff
@@ -0,0 +1,20 @@
...
+from .serializers import ChainSerializer
+
+
+class ChainsListView(ListAPIView):
+    serializer_class = ChainSerializer
+    pagination_class = LimitOffsetPagination
+    pagination_class.max_limit = 10
+    pagination_class.default_limit = 10
+
+    @swagger_auto_schema()
```
Uxio0, line 15: Do you need to do this to get the schema? It should be automatic without overriding `get`

src/config/urls.py
```diff
@@ -12,6 +12,7 @@
 
 urlpatterns = [
     path("api/v1/", include("safe_apps.urls", namespace="v1")),
+    path("api/v1/", include("chains.urls", namespace="chains")),
```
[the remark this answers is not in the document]
fmrsabino: Ah the branch got automatically rebased with the merge of the other one. This should only contain the commit `
7f07cab`.

Regarding your comment that change is part of #90

src/chains/models.py
fmrsabino: The integer is constrained to 4 bytes. I don't think chain ids have this limit though so I went with a string representation.

src/config/urls.py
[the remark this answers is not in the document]
Uxio0: I will comment there 🙂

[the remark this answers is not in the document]
fmrsabino: Also updated this PR to have only the changes of the pagination.

It had other changes because it was on top of the PR that just got merged.

src/chains/models.py
Uxio0: I would go with either a `IntegerField`, or if you are concerned about the size `Uint256Field`

`from gnosis.eth.django.models import EthereumAddressField, Uint256Field`

src/chains/tests/test_views.py
```diff
@@ -7,32 +7,93 @@
...
-                    "symbol": chain.currency_symbol,
-                    "decimals": chain.currency_decimals,
-                },
-            }
-        ]
+        json_response = {
+            "count": 1,
+            "next": None,
+            "previous": None,
+            "results": [
```
jpalvarezl, line 25: Happy that we have the same structure across all services

src/chains/models.py
fmrsabino: Since this affects the model (introduced before pagination) I will tackle it in a different PR. Is that ok? 🙂

PR #94, opened 2021-06-02
Use PositiveBigIntegerField to represent a chainId

fmrsabino:
- Use `PositiveBigIntegerField` (`0` to `9223372036854775807`) to represent a `chainId` (instead of a `CharField`) – https://docs.djangoproject.com/en/3.2/ref/models/fields/#positivebigintegerfield

### Open Questions

- The Django Rest Framework maps a `PositiveBigIntegerField` to a String. I've checked their model mapping and they don't seem to have a mapping for it so it is mapped as a string: https://github.com/encode/django-rest-framework/blob/d2994e0596c3163ac970b29dad6a61485f938045/rest_framework/serializers.py#L839-L864


Uxio0:
Are you sure? https://github.com/encode/django-rest-framework/blob/d2994e0596c3163ac970b29dad6a61485f938045/rest_framework/serializers.py#L855

Uxio0:
I was not aware of that, nice catch!

Uxio0 (approved):

fmrsabino:
> I was not aware of that, nice catch!

Seems to have been added to `3.1` 🙌 Serialisation is still as a string but I think that's ok since we restrict the type on the model

PR #108, opened 2021-06-15
Use nested namespaces for versioning

### LINKED ISSUE #107: Improve versioning namespace
Currently requests don't have a version set. This can be done for example with `URLPathVersioning` or `NamespaceVersioning`.

`NamespaceVersioning` seems to integrate quite well with `drf-yasg`

fmrsabino:
Closes #107 

- Url paths that require versioning are now under a nested namespace: `v1`
- This means that `chains` and `safe-apps` apps are now under `v1:chains` and `v1:safe-apps` respectively 
- Enable request versioning with `rest_framework.versioning.NamespaceVersioning`

### DIFF
--- a/src/chains/tests/test_views.py
+++ b/src/chains/tests/test_views.py
@@ -6,7 +6,7 @@ from .factories import ChainFactory
 
 class EmptyChainsListViewTests(APITestCase):
     def test_empty_chains(self):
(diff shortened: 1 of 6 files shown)


Uxio0 (approved):
Well done!

PR #110, opened 2021-06-16
Add trailing slash to chains/<id>/ endpoint

fmrsabino:
- `/api/v1/chains/<id>/` should now return the correct response – `200` if the `id` exists, instead of a `404`
- `/api/v1/chains/<id>` (no trailing slash) now redirects (`HTTP 301`) to the endpoint with the trailing slash. The configuration that enables this behaviour is `APPEND_SLASH` which is enabled by default

### DIFF
--- a/src/chains/urls.py
+++ b/src/chains/urls.py
@@ -6,5 +6,5 @@ app_name = "chains"
 
 urlpatterns = [
     path("", ChainsListView.as_view(), name="list"),
-    path("<pk>", ChainsDetailView.as_view(), name="detail"),
+    path("<pk>/", ChainsDetailView.as_view(), name="detail"),
 ]


src/chains/urls.py
```diff
@@ -6,5 +6,5 @@
 
 urlpatterns = [
     path("", ChainsListView.as_view(), name="list"),
```
Uxio0, line 8: It should be `"/"`

Uxio0 (changes requested):

src/chains/urls.py
fmrsabino: I've tried that 😅 It gives the following warning:

>?: (urls.W002) Your URL pattern '/' [name='list'] has a route beginning with a '/'. Remove this slash as it is unnecessary. If this pattern is targeted in an include(), ensure the include() pattern has a trailing '/'.

fmrsabino: for `chains/` it's the same behaviour as I described in the PR:

- `/chains` redirects (`HTTP 301`)
- `/chains/` works

Uxio0: My mistake, I've just checked it and was getting the same error, sorry

Uxio0 (approved):

PR #112, opened 2021-06-17
Add theme information to /chain

### LINKED ISSUE #86: Add theme information to the chain endpoint
Frontend clients have a theming option for each L2 chain. This information should be stored and retrieved with `/api/v1/chains` so that clients can render the respective theme for each chain.

The following should be added to the `/api/v1/chains` endpoint.

### Initial proposal:

```javascript
{
  ...
  "theme": {
       "light": {
           "textColor": <hex-color-string>,
           "backgroundColor": <hex-color-string>
       },
       "dark": {
           "textColor": <hex-color-string>,
           "backgroundColor": <hex-color-string>
       }
   }
}
```

### Update on the payload: https://github.com/gnosis/safe-config-service/issues/86#issuecomment-862340184

fmrsabino:
Closes #86 

- Add two new fields to `/chains`: `theme_text_color` and `theme_background_color`
- Both colours should be represented as an hexadecimal color (support for alpha bits is also present)
- These fields are stored as a `CharField` in the database. However in order to save them the following validations are done:
  * It should match the following regex: `^#(?:[0-9a-fA-F]{3,4}){1,2}$`
  * It should be maximum 9 characters – even though this is already covered by the regex expressions this is still required by the `CharField`


src/chains/tests/test_models.py
```diff
@@ -10,3 +13,28 @@ def test_str_method_outputs_name_chain_id(self):
             str(chain),
             f"{chain.name} | chain_id={chain.id}",
         )
+
+
+class ChainColorValidationTestCase(TransactionTestCase):
+    def test_invalid_text_colors(self):
+        param_list = ["aaa", "bbb", "#fffffffff", "zzz", "010", "", "a word", "#hhh"]
+        for invalid_color in param_list:
+            with self.subTest(msg=f"Invalid color {invalid_color} should throw"):
```
fmrsabino, line 22: `subTest` will generate a new test for each param in `param_list`. This had some problems while running with a normal `TestCase` with the rollbacks (probably due to the way the context of the `subTest` is set). Running with a `TransactionTestCase` works fine

src/chains/tests/test_models.py
```diff
@@ -10,3 +13,28 @@ def test_str_method_outputs_name_chain_id(self):
...
+                    )
+                ):
+                    chain = ChainFactory.create(theme_text_color=invalid_color)
+                    # run validators
+                    chain.full_clean()
+
+    def test_valid_text_colors(self):
+        param_list = ["#000", "#fff", "#00000000", "#ffffffff"] + [
+            Faker().hex_color() for _ in range(20)
+        ]
```
fmrsabino, line 36: Added some valid edge cases + some additional random 20 ones that should pass the test

src/chains/models.py
```diff
@@ -10,6 +17,12 @@ class Chain(models.Model):
     currency_symbol = models.CharField(max_length=255)
     currency_decimals = models.IntegerField(default=18)
     transaction_service_url = models.URLField(null=True)
+    theme_text_color = models.CharField(
+        validators=[color_validator], max_length=9, default="#fff"
+    )
+    theme_background_color = models.CharField(
+        validators=[color_validator], max_length=9, default="#000"
+    )
```
fmrsabino, line 25: @Uxio0 I thought about making a custom field for the color and I was reading more about it. I think for the use case that we have right now we can have it as a `CharField` with a `RegexValidator` what do you think?

src/chains/tests/test_models.py
Uxio0: I didn't know about `subTest`, but maybe in this case `@pytest.mark.parametrize` is more intuitive, what do you think? I need to dig on `pytest` as previously I was using regular python testing and there are some interesting things

src/chains/models.py
Uxio0: I think text with validator will work perfect. I would define a new field if you see yourself using in many places. There are examples of custom fields in `gnosis-py`

Uxio0 (approved):

src/chains/tests/test_models.py
fmrsabino: Unfortunately `@pytest.mark.parametrize` seems to not be compatible with Django Test Cases. I was able to make it work with a function test with access to DB via `@pytest.mark.django_db`. So the following works:

```python
@pytest.mark.django_db
@pytest.mark.parametrize
def test_invalid_text_colors(self):
```

But then I loose the assertion functions (like `assertRaises`). 😔

src/chains/models.py
fmrsabino: Sounds good 👍 If we end up using this in more places then I will consider adding a new field

src/chains/tests/test_models.py
Uxio0: Good to know 👍🏻

PR #124, opened 2021-06-28
Add relevance field to /chains

### LINKED ISSUE #123: Explicitly set /chains sorting
The chains returned by the `/chains` endpoint should allow the option to sort them by a relevance field (set manually)

fmrsabino:
Closes #123 

- Add a `relevance` field to `Chain`
- Sort fields according to `relevance` (ascending order). If relevance Is the same sort by the chain id


src/chains/admin.py
```diff
@@ -5,6 +5,14 @@
...
+        "id",
+        "name",
+        "rpc_url",
+        "relevance",
+    )
     search_fields = ("name", "id")
-    ordering = ("id",)
+    ordering = (
+        "relevance",
+        "id",
```
Uxio0, line 17: Maybe relevance, then name for ordering

src/chains/models.py
```diff
@@ -11,6 +11,9 @@
 
 class Chain(models.Model):
     id = models.PositiveBigIntegerField(verbose_name="Chain Id", primary_key=True)
+    relevance = models.SmallIntegerField(
+        default=100
+    )  # A lower number will indicate more relevance
```
Uxio0, line 16: Nice, that's the same that in the tx service

src/chains/views.py
```diff
@@ -10,7 +10,7 @@ class ChainsListView(ListAPIView):
     pagination_class = LimitOffsetPagination
     pagination_class.max_limit = 10
     pagination_class.default_limit = 10
-    queryset = Chain.objects.all()
+    queryset = Chain.objects.order_by("relevance", "id")
```
Uxio0, line 13: Don't do the sorting here: https://www.django-rest-framework.org/api-guide/filtering/#specifying-a-default-ordering

Uxio0: And maybe allow the users to order by name and relevance if they want

Uxio0 (changes requested):

src/chains/views.py
fmrsabino: Oh I saw this but since it added additional optional parameters for the clients to sort I decided to be more conservative. I can add this 👍

fmrsabino: Changed in 80ed735

Uxio0 (approved):
Looks good!

PR #128, opened 2021-06-28
Change safe-app primary key

### LINKED ISSUE #121: Use a generated unique id to represent a safe-app instead of a URL
Using the URL as a unique identifier for the safe-app is error prone as the same app can be represented in multiple forms (eg.: `https://example.com`, `https://example.com/`, `https://example.com/?`, `https://example.com/#` etc... 

We should then generate a unique identifier to represent each app instead of relying on the uniqueness of the url.

```javascript
GET /api/v1/safe-apps/

[
  {
    "id": <stringified-id>,
    ...
  }
]
```

fmrsabino:
Closes #121 

- Add `app_id` to `SafeApp` model
- `app_id` is the new primary key (previous primary key was the url)
- Remove `safe-apps` fixture – it does not correspond with the new model and the autogenerated primary key

#### Notes

In order to validate the schema before and after migration I ran `dumpdata` to verify the changes:

Before Migration:

```javascript
{
    "model": "safe_apps.safeapp",
    "pk": "https://app.aave.com/",
    "fields": {
        "name": "Aave v2",
        "icon_url": "https://app.aave.com/aave.svg",
        "description": "Non-custodial liquidity protocol to earn interest on deposits and borrow",
        "chain_ids": "[\"1\"]",
        "provider": null
    }
}
```

After Migration:

```javascript
{
    "model": "safe_apps.safeapp",
    "pk": 1,
    "fields": {
        "url": "https://app.aave.com/",
        "name": "Aave v2",
        "icon_url": "https://app.aave.com/aave.svg",
        "description": "Non-custodial liquidity protocol to earn interest on deposits and borrow",
        "chain_ids": "[\"1\"]",
        "provider": null
    }
}
```


src/safe_apps/models.py
```diff
@@ -11,7 +11,8 @@ def __str__(self):
 
 
 class SafeApp(models.Model):
-    url = models.URLField(primary_key=True)
+    app_id = models.BigAutoField(primary_key=True)
```
fmrsabino, line 14: FYI: Having just `id` as the name of the field conflicts with some internals regarding migration validation.

Adding a field `id` to the model results on an error stating that the field already exists (internal to Django).
Creating a custom migration to alter the field `id` to be the primary key results on an error stating that `id` does not exist.

One possible solution would be to copy the data present on the safe-apps table to a new table that sets this new `id`. But I find this solution easier to read and maintain.

fmrsabino:
@Uxio0 Just noticed that this makes the initial data fixture invalid. 

Adjusting the fixture to have a primary key set by us can be strange given that this is automatic field.
Having the `pk` set to `null` and delegate the generation of the pk to the model can result in duplicate apps since a similar app might already exist.

Should we remove the fixture completely?

Uxio0:
Yes, I think we can remove the fixture

fmrsabino:
> Yes, I think we can remove the fixture

Done here 2aab132

Uxio0 (approved):

PR #143, opened 2021-07-07
Add ENS Registry Address to each chain

### LINKED ISSUE #119: Add ENS Registry to each chain
In case the ENS registry is present on the Chain, then its address should be returned:

```javascript
GET /api/v1/chains/

{
  ...
  "ensRegistryAddress": "<checksummed-address>" // optional
}
```

@gnosis/safe-android @gnosis/safe-ios @gnosis/safe-web

fmrsabino:
Closes #119 

- Add `ens_registry_address` to `Chain`
- The `ens_registry_address` can either be `None` or needs to be a checksummed Ethereum address (other values are not valid)
- Adds Web3 dependency (in order to validate the Ethereum addresses)


src/chains/models.py
```diff
@@ -36,6 +37,9 @@ class Chain(models.Model):
     )
     gas_price_oracle_url = models.URLField(blank=True, null=True)
     gas_price_oracle_parameter = models.CharField(blank=True, null=True, max_length=255)
+    ens_registry_address = models.CharField(
```
Uxio0, line 40: Use `EthereumAddressField` from `gnosis.eth.django.model` (gnosis-py)

Uxio0 (changes requested):

src/chains/models.py
fmrsabino: Using the `EthereumAddressField` now e3604bc

requirements-dev.txt
```diff
@@ -7,3 +7,4 @@ flake8==3.9.2
 isort==5.9.1
 pre-commit==2.13.0
 pytest-django==4.4.0
+web3==5.20.1
```
fmrsabino, line 10: Moved the `web3` dependency to the `requirements-dev.txt` as it is only being used directly with the unit tests (to generate addresses)

Uxio0: I like this, but gnosis-py requires web3 and will install it, so no difference 😉

Uxio0 (approved):

PR #146, opened 2021-07-08
Add gweiFactor to GasPriceOracle

fmrsabino:
- Gwei factor is now part of the `gasPriceOracle` payload
- Default value is `1` ie.: it is assumed that the returned values by the oracle are in Gwei
- The `DecimalField` is set with a limit of `19` digits where `9` of those can be decimal places:
  * `1 Gwei == 1_000_000_000.000_000_000 Wei`
  * `1 Gwei == 0.000_000_001 ETH`

```javascript
{
  ...
  "gasPriceOracle": { // optional
       "url": <url-string>,
       "gasParameter": <string> // eg.: safeLow | average | fast
       "gweiFactor": <stringified-number> // required – default="1.000000000"
   }
}

```

The term `gweiFactor` matches what the SafeReact app currently uses.


src/chains/models.py
```diff
@@ -36,6 +36,11 @@ class Chain(models.Model):
     )
     gas_price_oracle_url = models.URLField(blank=True, null=True)
     gas_price_oracle_parameter = models.CharField(blank=True, null=True, max_length=255)
+    gas_price_oracle_gwei_factor = models.PositiveBigIntegerField(
```
Uxio0, line 39: What happens if one place returns the `gasPrice` in a unit bigger than `gwei`? Shouldn't multiplier be a `float`?

fmrsabino: Yes good point 👍

Uxio0 (approved):

PR #157, opened 2021-07-13
Add support for fixed gas price

### LINKED ISSUE #154: Support for fixed gas price without oracle
For some chains a fixed gas price can be used (instead of relying on a gas price oracle).

On the safe-react, the `GasPrice` can be either given with `gasPrice | gasPriceOracle?` or `gasPrice? | gasPriceOracle`: 
https://github.com/gnosis/safe-react/blob/ae206843c86aa1663ca6db66d3662c1882d36991/src/config/networks/network.d.ts#L80-L89

Given this, if `gasPrice` is given then `gasPriceOracle` should be `null` and if `gasPriceOracle` is given then `gasPrice` should be null.

### Open Questions

1. Is the `gasPrice` considered to be in `gwei`
2. Is it correct to assume that only one can be provided and not both?

cc/ @gnosis/safe-web

fmrsabino:
Closes #154 

- Add support for a fixed gas price – `gasPriceFixed`
- Each `Chain` can now either have a fixed gas price or a gas oracle – it needs to have one of these but not both
- Fixed gas price is represented in `Wei`

Payload is as follows:

If oracle is set:
```javascript
{
  ...
  "gasPrice" {
    "type": "oracle",
    "url": <url_string>,
    "gasParameter": <param_name>,
    "gweiFactor": <multiplier_number>
  }
}
```

If fixed price is set:
```javascript
{
  ...
  "gasPrice" {
    "type": "fixed",
    "weiValue": <stringified-number>
  }
}

```


src/chains/serializers.py
```diff
@@ -1,18 +1,25 @@
...
     url = serializers.URLField(source="gas_price_oracle_url")
     gas_parameter = serializers.CharField(source="gas_price_oracle_parameter")
     gwei_factor = serializers.DecimalField(
         source="gas_price_oracle_gwei_factor", max_digits=19, decimal_places=9
     )
 
 
+class GasPriceFixedSerializer(serializers.Serializer):
+    type = serializers.ReadOnlyField(default="fixed")
+    value = serializers.CharField(source="gas_price_fixed")
```
fmrsabino, line 20: Should this be `weiValue` to better represent which unit this field represents?

Uxio0: Yes, better explicit than implicit :)

Uxio0 (changes requested):

src/chains/tests/test_models.py
```diff
@@ -31,7 +56,9 @@ def test_oracle_gas_parameter_with_null_url(self):
 
     def test_null_oracle_gas_parameter_with_url(self):
         chain = ChainFactory.create(
-            gas_price_oracle_url=self.faker.url(), gas_price_oracle_parameter=None
+            gas_price_oracle_url=self.faker.url(),
+            gas_price_oracle_parameter=None,
+            gas_price_fixed_wei=None,
```
jpalvarezl, line 61: These are outwardly exposed?

src/chains/tests/test_views.py
```diff
@@ -174,10 +171,9 @@ def test_match(self):
                 "text_color": chain.theme_text_color,
                 "background_color": chain.theme_background_color,
             },
-            "gas_price_oracle": {
-                "url": chain.gas_price_oracle_url,
-                "gas_parameter": chain.gas_price_oracle_parameter,
-                "gwei_factor": "{0:.9f}".format(chain.gas_price_oracle_gwei_factor),
+            "gas_price": {
+                "type": "fixed",
```
jpalvarezl, line 175: Does this imply this endpoint will be polymorphic?

src/chains/models.py
```diff
@@ -42,21 +42,33 @@ class Chain(models.Model):
...
     gas_price_oracle_gwei_factor = models.DecimalField(
         default=1,
         max_digits=19,
         decimal_places=9,
         verbose_name="Gwei multiplier factor",
         help_text="Factor required to reach the Gwei unit",
     )
+    gas_price_fixed_wei = Uint256Field(
+        verbose_name="Fixed gas price (wei)", blank=True, null=True
+    )
```
jpalvarezl, line 57: This is output as a string as you can see in the tests, right?

src/chains/tests/test_views.py
```diff
@@ -252,3 +219,69 @@ def test_null_ens_registry_address(self):
...
+
+    def test_fixed_gas_256_bit(self):
+        ChainFactory.create(
+            id=1,
+            gas_price_fixed_wei="115792089237316195423570985008687907853269984665640564039457584007913129639935",
+        )
+        url = reverse("v1:chains:detail", args=[1])
+        expected_oracle_json_payload = {
+            "type": "fixed",
+            "weiValue": "115792089237316195423570985008687907853269984665640564039457584007913129639935",
```
jpalvarezl, line 281: Ok cool I think I answered my own question above :D

src/chains/tests/test_models.py
fmrsabino: Yes. Depending on the type of information available it'll either expose the oracle data or the fixed price data. (this is more a representation of the DB table – we avoided creating a new table with foreign keys just for gas related data)

src/chains/tests/test_views.py
fmrsabino: Yes right now only two types available: `fixed` and `oracle`

src/chains/models.py
fmrsabino: Django generates a human readable string for each field by default (to show in the admin interface).

So `gas_price_oracle_gwei_factor` would be shown as `Gas Price Oracle Gwei Factor` (in the admin)
For `gas_price_fixed_wei` it could be confusing having it generated as: `Gas Price Fixed Wei` so `verbose_name` just overrides that default.

src/chains/tests/test_views.py
fmrsabino: You can see the payload example in the PR description: https://github.com/gnosis/safe-config-service/pull/157#issue-689243974

This is what the CGW should expect.

Uxio0 (approved):

PR #159, opened 2021-07-15
Oracle parameter is required if oracle is set

fmrsabino:
- When the oracle url is set then the oracle parameter should also be required

```javascript
{
  "type": "oracle",
  "url": <string>, // required
  "gasParameter": <string>, // required
  "gweiFactor": <stringified-number>, // optional
}
```


fmrsabino:
Draft until https://github.com/gnosis/safe-config-service/pull/158 is merged

Uxio0 (approved):

PR #160, opened 2021-07-15
Rename /chains url fields to uri

fmrsabino:
- In order to make the models and view more consistent with the core services, `url` fields were renamed to `uri`


fmrsabino:
@Uxio0 

4e7fccd just renames on the view level
2a7bec4 also renames on the model level (requires migration)

What's your preference on this? Do you think we should rename both in the view and model?

Uxio0:
I don't think that renaming on the model is needed, but it's ok if it makes code cleaner

Uxio0 (approved):

PR #169, opened 2021-07-20
Add safe_apps_rpc_uri field to chain model

mikheevm:
What this PR does:
- Add `safe_apps_rpc_uri` to the chain model and view


run
```diff
@@ -31,10 +31,10 @@ function manage {
 
   # We need to collectstatic before we run our tests.
   if [ "${1-''}" == "test" ]; then
-    cmd python manage.py collectstatic --no-input
+    cmd python src/manage.py collectstatic --no-input
   fi
 
-  cmd python manage.py "${@}"
+  cmd python src/manage.py "${@}"
```
fmrsabino, line 37: Good catch. I think this stopped working when the Docker entrypoint changed (the entrypoint was `/appsrc/`)

Would it be ok if you opened a PR just with this change? (can also include the `.gitignore`). It's more to separate the changes to the application from the changes of the setup/configuration.

src/chains/models.py
```diff
@@ -24,6 +24,7 @@ class Chain(models.Model):
     )  # A lower number will indicate more relevance
     name = models.CharField(verbose_name="Chain name", max_length=255)
     rpc_uri = models.URLField()
+    safe_apps_rpc_uri = models.URLField()
```
fmrsabino, line 27: So this would be a mandatory field right?

fmrsabino (commented):
Another thing that I realised is that the RPC url currently does not include the project id for security reasons.

So the clients take the RPC url and add the respective project id (it's assumed that it's an Infura url) – see for example `/api/v1/chains/1/` on the staging environment

src/chains/models.py
mikheevm: I'm not sure to be honest. On the frontend, it's a mandatory field. Maybe we can make it optional and use the main rpc URL as a fallback

mikheevm:
@fmrsabino maybe we can include it, but limit the usage by origin. I don't know how it works for mobile applications though

fmrsabino:
> @fmrsabino maybe we can include it, but limit the usage by origin. I don't know how it works for mobile applications though

Will sync with the BE team today regarding this topic.

run
mikheevm: Done https://github.com/gnosis/safe-config-service/pull/170

src/chains/migrations/0019_add_safe_apps_rpc.py
```diff
@@ -0,0 +1,31 @@
...
+                    ("NO_AUTHENTICATION", "No Authentication"),
+                ],
+                default="NO_AUTHENTICATION",
+                max_length=255,
+            ),
+        ),
+        migrations.AddField(
+            model_name="chain",
+            name="safe_apps_rpc_uri",
+            field=models.URLField(default=None),
```
fmrsabino, line 28: The default value here is `None` which breaks the non-nullability constraints set by the model 🤔

src/chains/serializers.py
```diff
@@ -33,14 +33,25 @@ class CurrencySerializer(serializers.Serializer):
...
+            "authentication_source", "rpc_authentication"
+        )
+        uri_source = kwargs.pop("uri_source", "rpc_uri")
+        self.fields.update(
+            {
+                "authentication": serializers.CharField(source=authentication_source),
+                "value": serializers.URLField(source=uri_source),
+            }
+        )
+        super(RpcUriSerializer, self).__init__(*args, **kwargs)
```
fmrsabino, line 47: I have a suggestion regarding this `__init__`. I think one advantage of declaring the fields in the serialiser is that it's easier to read what is actually returned on the json level. Therefore, what do you think of the following? `BaseRpcUriSerializer` is the base class which requires both `get_authentication` and `get_uri` to be implemented.

```python
class BaseRpcUriSerializer(serializers.Serializer):
    authentication = serializers.SerializerMethodField()
    uri = serializers.SerializerMethodField()

    @abstractmethod
    def get_authentication(self, obj):
        pass

    @abstractmethod
    def get_uri(self, obj):
        pass


class RpcUriSerializer(BaseRpcUriSerializer):

    def get_authentication(self, obj):
        return obj.rpc_authentication

    def get_uri(self, obj):
        return obj.rpc_uri


class SafeAppsRpcUriSerializer(BaseRpcUriSerializer):

    def get_authentication(self, obj):
        return obj.safe_apps_rpc_authentication

    def get_uri(self, obj):
        return obj.safe_apps_rpc_uri
```

fmrsabino: Te declaration on the root object would then have the following:

```python
    @staticmethod
    @swagger_serializer_method(serializer_or_field=BaseRpcUriSerializer)
    def get_safe_apps_rpc_uri(obj):
        return SafeAppsRpcUriSerializer(obj).data

    @staticmethod
    @swagger_serializer_method(serializer_or_field=BaseRpcUriSerializer)
    def get_rpc_uri(obj):
        return RpcUriSerializer(obj).data
```

src/chains/migrations/0019_add_safe_apps_rpc.py
fmrsabino: Maybe we can set the default to the value of `rpc_uri`. The same for authentication

src/chains/serializers.py
mikheevm: Thanks! Nice idea

src/chains/models.py
fmrsabino: One idea could be to make it optional and when serialising it returns the `rpc_uri` if this is not set. What do you think? (same for authentication).

But if `safe_apps_rpc_uri` or `safe_apps_rpc_authentication` is set then both should exist in the model.

This can be done in a following PR if we agree to this

src/chains/tests/test_views.py
```diff
@@ -159,6 +167,7 @@ def test_no_match(self):
         self.assertEqual(response.status_code, 404)
 
     def test_match(self):
+        self.maxDiff = None
```
fmrsabino, line 170: This should be removed

fmrsabino (changes requested):

src/chains/migrations/0019_add_safe_apps_rpc.py
```diff
@@ -0,0 +1,44 @@
...
+from django.db.models import F
+
+
+def copy_rpc_fields(apps, schema_editor):
+    chain_model = apps.get_model("chains", "chain")
+    db_alias = schema_editor.connection.alias
+    chain_model.objects.using(db_alias).all().update(
+        safe_apps_rpc_uri=F("rpc_uri"),
+        safe_apps_rpc_authentication=F("rpc_authentication"),
+    )
```
fmrsabino, line 13: ⚠️ Currently this is a not reversible migration. Revert implementation should be simple. Check: https://docs.djangoproject.com/en/3.2/ref/migration-operations/#runpython

fmrsabino (changes requested):

src/chains/serializers.py
```diff
@@ -32,15 +34,40 @@ class CurrencySerializer(serializers.Serializer):
...
+    authentication = serializers.SerializerMethodField()
+    value = serializers.SerializerMethodField()
+
+    @abstractmethod
+    def get_authentication(self, obj):
+        pass
+
+    @abstractmethod
+    def get_value(self, obj):
+        pass
```
fmrsabino, line 47:
```suggestion
    @abstractmethod
    def get_authentication(self, obj):  # pragma: no cover
        pass

    @abstractmethod
    def get_value(self, obj):  # pragma: no cover
        pass
```

fmrsabino (changes requested):
Let's signal the coverage framework that the base abstract methods don't need to be tested.

fmrsabino (approved):
🚀 🚀 🚀

fmrsabino (approved):

PR #172, opened 2021-07-21
Add Chain RPC Authentication parameter

### LINKED ISSUE #171: Support for non-authenticated RPC urls
Currently the RPC url being used assumes that Infura is the provider. Under this assumption, the clients just append the Infura project ID to use that RPC url.

Given that there might be chains not available on Infura and that might not require authentication we can have two different types:

```javascript
GET /api/v1/chains/<chain-id>

{
  "rpcUri": {
    "type": <string>, // INFURA | OTHER
    "value": <url> 
  }
}
```

Other types can be added in the future.

cc/ @gnosis/safe-android @gnosis/safe-ios @gnosis/safe-web 

### Payload Update 1 https://github.com/gnosis/safe-config-service/issues/171#issuecomment-884099079

fmrsabino:
Closes #171 

- Chain RPCs now expose an authentication parameter so that clients know how to authenticate against this specific RPC 

```javascript
GET /api/v1/chains/<chain-id>
{
  ...
  "rpcUri": {
    "authentication": <API_KEY_PATH | NO_AUTHENTICATION>
    "value": <url> 
  }
}
```


src/chains/models.py
```diff
@@ -18,11 +18,18 @@
 
 
 class Chain(models.Model):
+    class RpcAuthentication(models.TextChoices):
+        API_KEY_PATH = "API_KEY_PATH"
+        NO_AUTHENTICATION = "NO_AUTHENTICATION"
+
```
fmrsabino, line 24: @jpalvarezl what's our standard for exposing multiple words as a value (for type association)?

mikheevm (approved):
LGTM

PR #176, opened 2021-07-26
Add block explorer uri templates

fmrsabino:
- Add Block Explorer URI templates for `address` and `tx_hash`
- This provides the support required for https://github.com/gnosis/safe-client-gateway/issues/515


src/chains/models.py
```diff
@@ -37,7 +37,11 @@ class RpcAuthentication(models.TextChoices):
         default=RpcAuthentication.NO_AUTHENTICATION,
     )
     safe_apps_rpc_uri = models.URLField(default="")
-    block_explorer_uri = models.URLField()
+    block_explorer_uri = (
+        models.URLField()
+    )  # TODO: Remove in a later migration (once the clients no longer need it)
```
jpalvarezl, line 42: Couldn't we technically already remove this? Technically this would break the GW  but, we could open the PR there too and merge more or less at the same time to minimize down time. As this is only affecting staging anyhow.

src/chains/tests/test_views.py
```diff
@@ -181,6 +189,10 @@ def test_match(self):
                 "value": chain.safe_apps_rpc_uri,
             },
             "block_explorer_uri": chain.block_explorer_uri,
+            "block_explorer_uri_template": {
+                "address": chain.block_explorer_uri_address_template,
+                "tx_hash": chain.block_explorer_uri_tx_hash_template,
```
jpalvarezl, line 194: I am wondering what would be the best way express this. The assumption is that there is always a template for `address` and `tx_hash` ? Introducing new types here would be a bit cumbersome, right? This looks a lot like a `HashMap<String, String>`

jpalvarezl: Maybe it would be better to have a `List<UriTemplate>` where `UriTemplate` is: 
```rust
struct UriTemplate {
  type: String
  uri: String
}
```
Sorry if this was already discussed with the clients.

src/chains/models.py
fmrsabino: I agree. This request came from the clients though after staging broke last time (which I think it's reasonable) and since the field name was different I decided to remove it later.

src/chains/tests/test_views.py
fmrsabino: For simplicity reasons each template is stored in a column. Giving the example that a new chain is added via the Admin interface how would you see a new template being added? 

Currently it is quite loose... it might not even have any template in the URL although this can be changed: I can add a validator that detects if the field being save has something like `{{template}}`

fmrsabino: > Maybe it would be better to have a `List<UriTemplate>` where `UriTemplate` is:
> 
> ```rust
> struct UriTemplate {
>   type: String
>   uri: String
> }
> ```
> 
> Sorry if this was already discussed with the clients.

The problem with this is that it does not cover situations where there might be multiple template fields to be replaced 🤔

mikheevm: To me, it seems the map serves well for this case, what advantages would a list have?

jpalvarezl: The advantage of using `UriTemplate` is typing. Using simply `HashMap<String, String>` makes me feel a bit insecure as to how clients would differentiate between the entry for `tx_hash` and the entry for `address` . It is just string comparison? That seems a bit error prone. Not sure how strict we need to be here, but I would say types make for better self-documenting code.

jpalvarezl: Admittedly, introducing polymorphism here, would make it so that adding a new type, is a breaking change of the API whereas a new entry in the `HashMap<String, String>` is easy. I would just be concerned about the time we need to `HashMap<(String, String, ..., String) , String>` and so on

mikheevm: Is it possible to define constants type for keys in the hashmap?

jpalvarezl: That could also be an alternative to have the keys as an enum, but then the variants in the enum would also have to match the string for the path variable name here: `{{placeholder}}`

jpalvarezl (approved):
After discussing at length alternative solutions, it seems like the current one covers entirely the expectations of the clients while managing adequately cost/benefit factors

PR #177, opened 2021-07-26
Chore: Set execution permission on docker entrypoint

mikheevm:
The current `chmod +x` command inside the Docker file doesn't set correct permissions because it seems that the `COPY` command is executed as `root` and thus `root` is the owner of this file, but the command is executed as a regular user. This is why one had to set correct permission before building the image (`COPY` copies with permissions). We can fix this by setting `chmod +x` on the file and pushing it to git because git captures permissions. We could also execute `chmod` as root but not sure if it's a better approach

Source: https://serverfault.com/questions/967580/chmod-changing-permissions-of-myscript-sh-operation-not-permitted


Uxio0 (approved):
I don't see any issues by pushing to git with run permissions

fmrsabino (approved):

PR #180, opened 2021-07-30
Add CGW Chains webhook

### LINKED ISSUE #173: Register CGW hook in order to invalidate its cache
A service hook should be implemented so that when Chain related data is updated, the CGW is notified about those changes and invalidates its cache

fmrsabino:
Closes #173 

- Client Gateway hook implementation – this hook is triggered when a `Chain` is updated, created or deleted
- Add two new environment variables: one for the client gateway url (`CGW_URL`) and another one for the flush token (`CGW_FLUSH_TOKEN`)

### DIFF
--- a/src/chains/signals.py
+++ b/src/chains/signals.py
@@ -1,0 +1,44 @@
+import logging
+from functools import cache
+from urllib.parse import urljoin
+
+import requests
+from django.conf import settings
+from django.db.models.signals import post_delete, post_save
+from django.dispatch import receiver
+
+from .models import Chain
+
+logger = logging.getLogger(__name__)
+
+
+@cache
+def setup_session():
+    session = requests.Session()
+    adapter = requests.adapters.HTTPAdapter(max_retries=3)
+    session.mount("http://", adapter)
+    session.mount("https://", adapter)
+    return session
+
+
+@receiver(post_save, sender=Chain)
+@receiver(post_delete, sender=Chain)
+def on_chain_update(sender, **kwargs):
+    logger.info("Chain update. Triggering CGW webhook")
+    cgw_url = settings.CGW_URL
+    if cgw_url is None:
+        logger.error("CGW_URL is not set. Skipping hook call")
+        return
+    cgw_flush_token = settings.CGW_FLUSH_TOKEN
+    if cgw_flush_token is None:
+        logger.error("CGW_FLUSH_TOKEN is not set. Skipping hook call")
+        return
+
+    url = urljoin(cgw_url, f"/v1/flush/{cgw_flush_token}")
+    try:
+        post = setup_session().post(
+            url, json={"invalidate": "Chains"}
+        )  # TODO: if this throws it might leak the token to the logger
+        post.raise_for_status()
+    except Exception as error:
+        logger.error(error)
--- a/src/chains/signals.py
+++ b/src/chains/signals.py
@@ -36,9 +36,7 @@ def on_chain_update(sender, **kwargs):
 
     url = urljoin(cgw_url, f"/v1/flush/{cgw_flush_token}")
     try:
-        post = setup_session().post(
-            url, json={"invalidate": "Chains"}
-        )  # TODO: if this throws it might leak the token to the logger
+        post = setup_session().post(url, json={"invalidate": "Chains"})
         post.raise_for_status()
     except Exception as error:
         logger.error(error)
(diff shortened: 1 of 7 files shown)


src/chains/signals.py
```diff
@@ -0,0 +1,45 @@
...
+    cgw_flush_token = os.environ.get("CGW_FLUSH_TOKEN")
+    if cgw_flush_token is None:
+        logger.error("CGW_FLUSH_TOKEN is not set. Skipping hook call")
+        return
+
+    url = urljoin(cgw_cluster_url, f"/v1/flush/{cgw_flush_token}")
+    try:
+        post = session.post(
+            url, json={"invalidate": "Chains"}
+        )  # TODO: if this throws it might leak the token to the logger
```
fmrsabino, line 42: Since the token is part of the URL I wonder if that happens anyways with our tracking and this is a non-issue

Uxio0: Sorry, I don't understand what you mean 😅

src/chains/tests/test_signals.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+    def test_on_chain_delete_hook_call(self, mock_post):
+        chain = ChainFactory.create()
+
+        chain.delete()
+
+        # Call count should be 2: one for creation and one for deletion
+        self.assertEqual(mock_post.call_count, 2)
+
+    @patch("requests.sessions.Session.post")
```
Uxio0, line 46: Why don't you use here `responses` library?

src/chains/signals.py
```diff
@@ -0,0 +1,45 @@
...
+
+def setup_session():
+    session = requests.Session()
+    adapter = requests.adapters.HTTPAdapter(max_retries=3)
+    session.mount("http://", adapter)
+    session.mount("https://", adapter)
+    return session
+
+
+session = setup_session()
```
Uxio0, line 22: Instead of having a global variable you can add the https://docs.python.org/3/library/functools.html#functools.cache annotation to the function, yous FYI, but I think it's good this way too 😉

src/chains/signals.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@receiver(post_save, sender=Chain)
+@receiver(post_delete, sender=Chain)
+def on_chain_update(sender, **kwargs):
+    logger.info("Chain update. Triggering CGW webhook")
+    cgw_cluster_url = os.environ.get("CGW_URL")
```
Uxio0, line 29: Don't use `os.environ` in a Django application. Load the environment variables on the configuration and here import `django.conf.settings` to use them. This way you prevent runtime errors

Uxio0 (changes requested):

src/chains/signals.py
fmrsabino: If the `POST` call is unsuccessful then this is going to log an exception like:


>[WARNING] [MainProcess] Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x107580610>: Failed to establish a new connection: [Errno 61] Connection refused')': /v1/flush/some_random_token

Here `some_random_token` is present in the logs (in production this would be the real token). But since it's also part of the path it'd be already part of what is sent to Kibana 😅

Uxio0: Yeah, I think it's not a problem 👍🏻

src/chains/tests/test_signals.py
fmrsabino: I agree. One less mock.

fmrsabino:
@Uxio0 applied your suggestion in f1a7c99

I wanted to have a separate commit for it but I squashed it by mistake 🙈

src/chains/tests/test_signals.py
```diff
@@ -0,0 +1,92 @@
...
+        assert not mock_post.called
+
+    @patch("requests.sessions.Session.post")
+    @override_settings(
+        CGW_URL="http://127.0.0.1",
+    )
+    def test_on_chain_update_with_no_flush_token_set(self, mock_post):
+        ChainFactory.create()
+
+        assert not mock_post.called
```
fmrsabino, line 23: Maybe instead of mocking here I can do

```python
assert len(responses.calls) == 1
```

what do you think?

fmrsabino: Done here cbd3da4 (can revert if the previous way is preferable)

Uxio0 (approved):

PR #183, opened 2021-08-02
Check pending Django migrations in CI

fmrsabino:
- Add step to CI to check if there are any migrations that should've been created and were not created (without creating them)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -85,6 +85,8 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -r requirements-dev.txt
+      - name: Check pending migrations
+        run: python src/manage.py makemigrations --check --dry-run
       - name: Run migrations
         run: python src/manage.py migrate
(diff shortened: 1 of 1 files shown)


fmrsabino:
@Uxio0 the CI is failing as expected. A missing migration was neither created nor applied.

This is fixed here: https://github.com/gnosis/safe-config-service/pull/184

Uxio0:
We already talked about it when setting up the CI: `python manage.py check` 😉

Uxio0:
I didn't know about this:

```
      - name: Check pending migrations
        run: python src/manage.py makemigrations --check --dry-run
```

I will add it asap on tx service CI too

Uxio0 (approved):
Nice!

fmrsabino:
@Uxio0 Rebased to `main` after applying the missing migration. So the correct CI output should now be green 🤞

Uxio0 (approved):

PR #192, opened 2021-08-09
Added new `ChainInfo` issue template

### LINKED ISSUE #191: Add issue template in github for requesting adding a new chain
The new issue template should be specific to adding new `ChainInfo` provide a list of the information that needs to be provided as well as all the configuration steps required to be performed by the backend team. 

This would be an interim solution, unless we see that it works well.

How web does it: https://docs.google.com/document/d/1k7X1oQNtl-WZFKIlydZJtdV0Wjf_qRsJ_9R56fryCL8/edit

jpalvarezl:
Closes #191 

- Transfered from CGW
- Checklist for steps for adding and configuring a new chain

### DIFF
--- a/.github/ISSUE_TEMPLATE/new_chain_request.md
+++ b/.github/ISSUE_TEMPLATE/new_chain_request.md
@@ -1,0 +1,80 @@
+---
+name: Chain support request
+about: Add support to a new chain
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+## New chain support request
+
+Please tick off the boxes as the steps in the process are completed.
+
+## Chain Metadata
+
+We need to collect the metadata for a chain. The following section contains a sample.
+
+<details>
+<summary>Example fields</summary>
+
+```json
+{
+    "chainId": "1",
+    "chainName": "Mainnet",
+    "rpcUri": {
+    "authentication": "API_KEY_PATH",
+    "value": "https://mainnet.infura.io/v3/"
+    },
+    "safeAppsRpcUri": {
+    "authentication": "API_KEY_PATH",
+    "value": "https://mainnet.infura.io/v3/"
+    },
+    "blockExplorerUriTemplate": {
+    "address": "https://etherscan.io/address/{{address}}",
+    "txHash": "https://etherscan.io/tx/{{txHash}}"
+    },
+    "nativeCurrency": {
+    "name": "Ether",
+    "symbol": "ETH",
+    "decimals": 18,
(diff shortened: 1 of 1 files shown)


fmrsabino (commented):
In general I think it looks good but I'd say that we should make it simpler to create a request to add a new chain. 

I think for example we should avoid using Json for making this request as it can be quite cryptic on the meaning of some fields (eg.: `authentication": "API_KEY_PATH"`)

My recommendation is the following:
1) Make a simple form out of it. Maybe we can try this out https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms#about-yaml-syntax-for-issue-forms
2) If 1 is not possible then maybe just simple text (like question and answer) would be better in my opinion than json. This allows us to further expand what each field means

jpalvarezl:
@fmrsabino do you know how this could be checked? I don't think I can easily test `issue forms` locally. Also, they are a beta feature, per the link you attached. Are you sure you want this included in the repository?

fmrsabino:
> @fmrsabino do you know how this could be checked? I don't think I can easily test `issue forms` locally. Also, they are a beta feature, per the link you attached. Are you sure you want this included in the repository?

We don't need to include it. Was more a suggestion that I saw since I did not know that they had this forms in templates. We can go without it (no strong opinion here).

> It would be great to get your input on what fields we should request in the template. I envisioned this initially as a check list of configuration items for our side, but if we need to request fields from people, maybe we should give more thought as to what we can expect from someone requesting a new network.

✅ – should be provided in the request
❌ – should be provided by us
❓ – open questions/optional

> I would say:
> 
* chainId ✅
* chainName ✅
* rpc node url ❌ – this should be decided on our side (infrastructure wise)
* safe app rpc node url (optional) ❌ – same as above
* block explorer url (optionally provide the values for `tx_hash` and `address` template) ✅
* nativeCurrency: name, symbol, decimals and logo url ✅
* text & background colour ❌ – since this is mostly for our clients I think it'd be up for our designers/product to decide on this one.


> It would be up to us to fill the rest of the details. Namely:
> 
* tx service base url ❌ (on us as you suggested)
* gasPrice object ❓ – maybe we can have this as a suggestion – maybe an oracle can be suggested in the request or if the gas price is fixed this should also be provided
* ens registry address ✅ – the request should probably include the ENS registry address if available (so that we have a reference of the deployment)
* recommended Master copy ❌ (on us)

What do you think?

PR #197, opened 2021-08-20
Set publish workflows to depend on tests

### LINKED ISSUE #196: Trigger deployment after tests/style checks
Currently the deployment to staging and production happens in parallel with tests/style checks. This is usually not a problem because the `main` branch is protected and the test checks are required to merge a PR.

However the following scenario can happen:
- Create branch
- Push failing changes
- Tag branch (this triggers both test checks and a docker deploy) – tests will fail but the docker deploy can succeed and a failed tagged version is then published on DockerHub

`workflow_run` might be useful to tackle this: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run

fmrsabino:
Closes #196 

- The publish workflow happens in parallel to the tests/health checks.
- This is generally ok since the publishing usually happens from the `main` branch OR tagged commits on the `main` branch
- However, tagged commits can happen on any branch which would still run the tests/health check but since they happen in parallel, publishing might be successful even if these checks fail
- `docker-publish-staging` and `docker-publish-release` now depend on `flake8, isort, black, django-check`


.github/workflows/test.yml
```diff
@@ -98,3 +106,90 @@ jobs:
...
+          cache-to: type=local,dest=/tmp/.buildx-cache-new
+      - # Temp fix
+        # https://github.com/docker/build-push-action/issues/252
+        # https://github.com/moby/buildkit/issues/1896
+        name: Move cache
+        run: |
+          rm -rf /tmp/.buildx-cache
+          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
+      - name: Image digest
+        run: echo ${{ steps.docker_build.outputs.digest }}
```
jpalvarezl, line 195: Would you be against merging the deployment jobs? It seems that they only differ in line `~184` and line `~140`. 

Something along the lines of this: https://stackoverflow.com/a/63622205

Where you can have something like:

```
- name: Set tag
        if: github.event_name == 'release' && github.event.action == 'released'
        run: echo 'DOCKER_TAG=${{ github.event.release.tag_name }}' >> "$GITHUB_ENV"
        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
        run: echo 'DOCKER_TAG=staging}' >> "$GITHUB_ENV"
```

fmrsabino: That would indeed look nicer. I will take a look into it.

.github/workflows/test.yml
```diff
@@ -129,6 +127,14 @@ jobs:
...
           username: ${{ secrets.DOCKER_USER }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Set tag (staging)
+        # Set "staging" as Docker tag if the event is a push to main branch
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        run: echo "DOCKER_TAG=staging" >> $GITHUB_ENV
+      - name: Set tag (release)
+        # Set the tag versions as Docker tag if it is a release event
+        if: github.event_name == 'release' && github.event.action == 'released'
+        run: echo "DOCKER_TAG=${{ github.event.release.tag_name }}" >> $GITHUB_ENV
```
fmrsabino, line 137: @jpalvarezl I think it's not possible to have multiple run conditions on a single `step`. See: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif

jpalvarezl: That's unfortunate! but this seems like the best we can do with what we have.

.github/workflows/test.yml
```diff
@@ -1,6 +1,14 @@
 name: "CI"
 
-on: push
+# The workflow should be triggered on any push and release events.
+# Release events can push tags (triggering a push event).
+# Therefore:
+#  - docker-publish-staging – is only triggered on push events to main branch
+#  - docker-publish-release – is only triggered on release events
+on:
+  push:
```
jpalvarezl, line 9: Will you skip `pull_request` for now? This would leave open again to the situation where inbound changes from a fork won't trigger the workflow, unless I am missing something.

jpalvarezl (approved):
LGTM

.github/workflows/test.yml
fmrsabino: Yes for now I'd not change this but this is something that we can definitely consider to add as a part of a separate issue 👍

PR #210, opened 2021-09-02
Add support for collection of oracles to chains endpoint

### LINKED ISSUE #209: Gas price oracles should return a list instead a single oracle
Currently we need some fallback mechanism in order to use an alternative gasPrice oracle in case the current one is failing. In order to achieve this we created a fallback system that accepts a list of oracles instead of a single oracle and so that way we can check next resource whenever we get an error from the previous one

```
"gasPrice": [{
    "type": "oracle",
    "uri": "https://www.gasnow.org/api/v3/gas/price?utm_source=:gnosis_safe",
    "gasParameter": "fast",
    "gweiFactor": "1",
  }, {
    "type": "oracle",
    "uri": "https://ethgasstation.info/json/ethgasAPI.json",
    "gasParameter": "fast",
    "gweiFactor": "100000000.000000000"
  }]
```

fmrsabino:
Closes #209 

- **Breaking Change `/api/v1/chains`**:  each chain is now allowed to have multiple gas price configurations (resulting in a collection instead of a single object of `fixed` and `oracle` gas prices).
- The new `GasPrice` for each chain is now ranked – a lower value means that it'd show up higher on the list (eg.: `rank 1 > rank 100`)
- The new `GasPrice` collection is allowed to be empty – this is not only to reduce complexity but also because the current relationship between `GasPrice <> Chain` means that a `Chain` needs to exist first before assigning a `GasPrice` to it.


src/chains/migrations/0023_create_gas_price_model.py
```diff
@@ -0,0 +1,75 @@
...
+import gnosis.eth.django.models
+from django.db import migrations, models
+
+
+def copy_gas_prices(apps, schema_editor):
+    GasPrice = apps.get_model("chains", "GasPrice")
+    Chain = apps.get_model("chains", "Chain")
+    db_alias = schema_editor.connection.alias
+
+    GasPrice.objects.using(db_alias).bulk_create(
```
Uxio0, line 13: Just curious, is this `using` needed?

src/chains/serializers.py
```diff
@@ -5,21 +5,35 @@
...
+        return GasPriceOracleSerializer(instance).data
+    elif instance.fixed_wei_value and instance.oracle_uri is None:
+        return GasPriceFixedSerializer(instance).data
+    else:
+        raise APIException(
+            f"The gas price oracle or a fixed gas price was not provided for chain {instance.chain}"
+        )
+
+
+class GasPriceSerializer(serializers.Serializer):
```
Uxio0, line 34: Nice done

Uxio0 (dismissed):

src/chains/migrations/0023_create_gas_price_model.py
fmrsabino: I would say that it is not required reading this: https://docs.djangoproject.com/en/3.2/ref/models/querysets/#using

Since we are using just one DB. I think we can remove it 👍

fmrsabino: Removed it here 60342d3

Uxio0 (approved):

PR #217, opened 2021-09-08
Add S3 bucket integration for currency logos

### LINKED ISSUE #204: Integrate with S3 bucket for native currency logos
Our S3 bucket currently stores logos using the following paths:
- `contracts/logos/{address}.png` (for contract addresses)
- `/{address}.png` (for tokens)

Since the native currency of a chain is not tied to a specific contract deployment we can use the chain information as the path for the logo being used: `/{chain}/logo.png`

@gnosis/safe-services

fmrsabino:
Closes #204 

- Add S3 storage integration using `django-storages`
- Native currency logos are stored in  the following path `safe-config/chains/{chain_id}_currency_logo.{ext}`
- Add the following environment variables:
  * `DEFAULT_FILE_STORAGE` (default is `S3Boto3Storage`)
  * `AWS_ACCESS_KEY_ID` (no default)
  * `AWS_SECRET_ACCESS_KEY` (no default)
  * `AWS_STORAGE_BUCKET_NAME` (no default)
  * `AWS_S3_CUSTOM_DOMAIN` (no default)
- If `S3Boto3Storage` is set, the AWS environment variables also need to be set (an exception is thrown otherwise)
- For local development `FileSystemStorage` can be used and files will be stored under `{repoRoot}/media/` (see `MEDIA_ROOT`)
- For testing purposes, `FileSystemStorage` is automatically set for all the tests and a temporary directory is created and removed on test execution


src/chains/models.py
```diff
@@ -17,6 +18,11 @@
 sem_ver_validator = RegexValidator(SEM_VER_REGEX, "Invalid version (semver)", "invalid")
 
 
+def native_currency_path(instance: "Chain", filename):
+    _, file_extension = os.path.splitext(filename)  # file_extension includes the dot
+    return f"safe-config/chains/{instance.id}_currency_logo{file_extension}"
```
Uxio0, line 23: Maybe change this path to `chains/{instance.id}}/currency_logo{extension}`. I might make mandatory that extension is `.png` to prevent issues with the clients

Uxio0 (changes requested):

Uxio0 (approved):

PR #218, opened 2021-09-09
Add Pillow build dependencies to Dockerfile

fmrsabino:
Since the introduction of `Pillow` in da146ef, additional build dependencies are required which were not present in the `Dockerfile` – https://pillow.readthedocs.io/en/latest/installation.html#building-on-linux


Uxio0 (approved):
This kind of things made me migrate to the regular non alpine build 😉

PR #222, opened 2021-09-13
Bump service version to 2.2.0

fmrsabino:
Changelog: https://github.com/gnosis/safe-config-service/compare/v2.1.0...bump-service-2-2-0


fmrsabino:
Ready to review but I'm wondering if we can have this https://github.com/gnosis/safe-config-service/pull/221 included. It would decrease the size of the next production image

Uxio0 (approved):

PR #235, opened 2021-09-27
Add vpcTransactionService field to Chain

### LINKED ISSUE #132: Add internal transaction service url to /chains
In addition to the public transaction url returned for each chain, an internal one should be returned and should be only used by the client gateway. 

```javascript
GET /api/v1/chains/

[
  {
    ...
    "transactionService": <url>,
    "transactionServiceCluster": <url> // this should only be used by the client gateway and not exposed to the clients
  }
]
```

@gnosis/safe-services

fmrsabino:
Closes #132 

- Adds a new field to `Chain` – `vpc_transaction_service_uri`
- This field is similar to `transaction_service_uri` but will hold the VPC url of the transaction service


jpalvarezl (approved):

src/chains/migrations/0028_chain_vpc_transaction_service_uri.py
```diff
@@ -0,0 +1,19 @@
...
+    dependencies = [
+        ("chains", "0027_chain_description"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="chain",
+            name="vpc_transaction_service_uri",
+            field=models.URLField(default=""),
+            preserve_default=False,
```
jpalvarezl, line 17: Should we constrain this to be non-blank and non-nullable?

fmrsabino: `non-blank` and `non-nullable` are the defaults: `blank=False, null=False` for all Django Fields.

https://docs.djangoproject.com/en/3.2/ref/models/fields/#null
https://docs.djangoproject.com/en/3.2/ref/models/fields/#blank

fmrsabino:
@jpalvarezl Converted this one to draft just to target a new release (so we don't accidentally merge it)

PR #241, opened 2021-09-30
Add EIP-3770 support

### LINKED ISSUE #232: Add short name to chain info
Reference: https://github.com/gnosis/safe/issues/487

To support [EIP-3770](https://github.com/ethereum/EIPs/pull/3770) the client require the short name for each network. The short name is unique for each chain id and can be retrieved from [ethereum-lists/chains](https://github.com/ethereum-lists/chains) or https://chainid.network/shortNameMapping.json

fmrsabino:
Closes #232 

- Adds `shortName` to the chains endpoint – EIP-3770 support https://github.com/ethereum/EIPs/pull/3770

```javascript
GET /api/v1/chains

{
  "chainId": "1",
  "chainName": "Mainnet",
  "shortName": "eth",
  ...
}
```

- The migration will migrate any chain with the ids present on this list https://chainid.network/shortNameMapping.json (as of today)
- Introduces `TestMigrations` – this `TestCase` class allows writing tests that target migrations in an easy manner. An example is included for migration `0029` – `Migration0029TestCase`


src/chains/migrations/tests/test_migration_0029.py
```diff
@@ -0,0 +1,38 @@
...
+            currency_symbol="ETH",
+            currency_decimals=18,
+            currency_logo_uri="https://gnosis-safe-token-logos.s3.amazonaws.com/ethereum-eth-logo.png",
+            transaction_service_uri="http://mainnet-safe-transaction-web.safe.svc.cluster.local",
+            vpc_transaction_service_uri="",
+            theme_text_color="#001428",
+            theme_background_color="#E8E7E6",
+            ens_registry_address="0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e",
+            recommended_master_copy_version="1.3.0",
+        )
```
fmrsabino, line 31: Here I will add the other networks that we currently support in production

src/chains/models.py
```diff
@@ -33,6 +33,7 @@ class RpcAuthentication(models.TextChoices):
         default=100
     )  # A lower number will indicate more relevance
     name = models.CharField(verbose_name="Chain name", max_length=255)
+    short_name = models.CharField(verbose_name="EIP-3770 short name", max_length=255)
```
rmeissner, line 36: This could even be unique (and maybe should to avoid issues in the future)

rmeissner (dismissed):
Minor comment, besides that it looks good

src/chains/models.py
fmrsabino: I agree 👍 Good point. I will add it.

src/chains/migrations/tests/utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+# Use this TestCase class in order to test migrations
+# migrate_from – the name of the migration where the test starts
+# migrate_to – the target migration
+#
+# Example:
+# You can set models on a specific migration inside your implementation of setUpBeforeMigration()
+# The validation can then be performed on each test function of the Test class
+# This class is heavily inspired by https://www.caktusgroup.com/blog/2016/02/02/writing-unit-tests-django-migrations/
+class TestMigrations(TestCase):
```
Uxio0, line 15: Nice, this is really useful

Uxio0 (approved):
Aren't you adding the new field to django admin list view?

PR #289, opened 2021-11-02
Add Wallet model

fmrsabino:
This adds support for https://github.com/gnosis/safe-client-gateway/issues/674

- Introduce `Wallet` model – it contains just a `name` that uniquely represents a wallet
- `Wallet` can reference multiple `Chains` – this relationship represents that the given `Wallet` is supported by the given `Chains`.
- Adds a field `disabledWallets` to `/api/v1/chains` – this field is a JSON vector of strings which contain the names of the wallets **not supported** by a given chain
  * The reason to have **not supported** is because by default, we assume that all chains are supported
  * However this means that once a `Chain` gets added, the `Wallets` should be respectively updated to include that `Chain`

### Admin Panel Overview

Example:

image

The above example adds a `Wallet` named `metamask` to `Mainnet` and `Rinkeby`. This means that `Polygon` will have `disabledWallets = ["metamask"]`


src/chains/models.py
```diff
@@ -76,6 +77,12 @@ class RpcAuthentication(models.TextChoices):
         max_length=255, validators=[sem_ver_validator]
     )
 
+    def get_disabled_wallets(self) -> QuerySet["Wallet"]:
+        all_wallets = Wallet.objects.all()
+        enabled_wallets = self.wallet_set.all()
```
Uxio0, line 82: Why not using `Wallet.objects.exclude(pk__in=self.wallet_set.values('pk')))`?

Uxio0 (approved):

src/chains/models.py
fmrsabino: Good point. 

With `difference` on Chain `4` the following SQL is generated:

```SQL
(SELECT "chains_wallet"."name" FROM "chains_wallet") EXCEPT
    (SELECT "chains_wallet"."name" FROM "chains_wallet" 
    INNER JOIN "chains_wallet_chains" ON 
        ("chains_wallet"."name" = "chains_wallet_chains"."wallet_id") 
        WHERE "chains_wallet_chains"."chain_id" = 4)
````

With `Wallet.objects.exclude(pk__in=self.wallet_set.values('pk')))` the following is generated:

```SQL
SELECT "chains_wallet"."name" FROM "chains_wallet" WHERE NOT 
    ("chains_wallet"."name" IN 
        (SELECT U0."name" FROM "chains_wallet" U0 
        INNER JOIN "chains_wallet_chains" U1 ON 
        (U0."name" = U1."wallet_id") WHERE U1."chain_id" = 4))
```

Do you have a preference?

Uxio0: I _feel_ that the second one will be more optimal (usually the `UNION`, `EXCEPT`... are not very efficient), but I don't think it will be a big issue. You can always use `EXPLAIN` to check how the database resolves the query

PR #290, opened 2021-11-03
Add Block Explorer API template

fmrsabino:
Adds support for https://github.com/gnosis/safe-client-gateway/issues/673

- Add a new field to `blockExplorerUriTemplate` – `api`
- This new field represents an URL with the path to the `api` of the respective block explorer
- `blockExplorerUriTemplate.api` has the following components in the template:
  * `{{module}}` (eg.: `contract`)
  * `{{action}}` (eg.: `getAbi`)
  * `{{address}}` (Ethereum address used for action – if needed)
  * `{{apiKey}}` – API key for the block explorer (set by the clients) 

```javascript
GET /api/v1/chains

[
  {
    "chainId": <string>,
    "chainName": <string>,
    "blockExplorerUriTemplate": {
        "address": "https://etherscan.io/address/{{address}}", // example
        "txHash": "https://etherscan.io/tx/{{txHash}}" // example
        "api": "https://api.etherscan.io/api?module={{module}}&action={{action}}&address={{address}}&apiKey={{apiKey}}" // new field example
      },
  }
]
```


Uxio0 (dismissed):

fmrsabino:
@Uxio0 had to rebase

src/chains/migrations/0031_chain_block_explorer_uri_api_template.py
```diff
@@ -0,0 +1,53 @@
...
+
+API_TEMPLATE_MAPPING = {
+    1: f"https://api.etherscan.io/api{QUERY_PARAM_TEMPLATE}",
+    4: f"https://api-rinkeby.etherscan.io/api{QUERY_PARAM_TEMPLATE}",
+    56: f"https://api.bscscan.com/api{QUERY_PARAM_TEMPLATE}",
+    100: f"https://blockscout.com/poa/xdai/api{QUERY_PARAM_TEMPLATE}",
+    137: f"https://api.polygonscan.com/api{QUERY_PARAM_TEMPLATE}",
+    246: f"https://explorer.energyweb.org/api{QUERY_PARAM_TEMPLATE}",
+    42161: f"https://api.arbiscan.io/api{QUERY_PARAM_TEMPLATE}",
+    73799: f"https://volta-explorer.energyweb.org/api{QUERY_PARAM_TEMPLATE}",
```
katspaugh, line 52: These chainIds can potentially be featured in other migrations, maybe it would make sense to use constants? Just a suggestion, I realize it's just a migration.

Uxio0 (approved):

src/chains/migrations/tests/test_migration_0031.py
```diff
@@ -0,0 +1,28 @@
+from .utils import TestMigrations
+
+
+class Migration0029TestCase(TestMigrations):
+    migrate_from = "0030_wallet"
+    migrate_to = "0031_chain_block_explorer_uri_api_template"
+
+    fixtures = ["src/chains/migrations/tests/fixtures/0030_chains.json"]
+
+    def test_block_explorer_uri_api_template_is_set(self) -> None:
```
jpalvarezl, line 10: Just to be sure I completely understand what happens here:
- first you define `from` which and `to` which migration you go.
- The fixture contains pre-migration `json`
- The migration is applied
- Assert that the objects contain the fields according to changes post-migration?

jpalvarezl (approved):
Just one question, but everything seems good!

src/chains/migrations/tests/test_migration_0031.py
fmrsabino: That's exactly it 🙌 

Sidenote: Even though we have Factories to generate the models, these factories are tied to latest version of the model. So I took `dumpdata` from migration `0030` and applied the json (instead of creating each object one by one).

jpalvarezl: That makes sense, I suppose that using the factory makes more sense when testing handlers or business logic in general 🤔

fmrsabino: The `Factory` just creates an object and saves it in the DB so it can be used anywhere – but it is tied to the latest version of the model (which we don't want in this case).

PR #291, opened 2021-11-03
Add Feature model

fmrsabino:
Adds support for https://github.com/gnosis/safe-client-gateway/issues/675

- Add a new model – `Feature` 
- `Feature` represents features that are enabled for specific chains
- It is represented by a single `name` (its primary key)
- Add a migration that adds the following features – `CONTRACT_INTERACTION`, `DOMAIN_LOOKUP`, `ERC721`, `ERC1155`, `SAFE_APPS` and `SPENDING_LIMIT`


Uxio0 (dismissed):

src/chains/models.py
```diff
@@ -113,3 +121,25 @@ def clean(self) -> None:
...
+
+class Wallet(models.Model):
+    # A wallet can be part of multiple Chains and a Chain can have multiple Wallets
+    chains = models.ManyToManyField(
+        Chain, blank=True, help_text="Chains where this wallet is enabled."
+    )
+    name = models.CharField(primary_key=True, max_length=255)
+
+    def __str__(self) -> str:
+        return f"Wallet: {self.name}"
```
katspaugh, line 134: Something is wrong with this PR, wasn't Wallet already merged?

src/chains/migrations/0032_feature.py
```diff
@@ -0,0 +1,49 @@
...
+
+from ..models import Feature
+
+DEFAULT_FEATURES = [
+    Feature(name="CONTRACT_INTERACTION"),
+    Feature(name="DOMAIN_LOOKUP"),
+    Feature(name="ERC721"),
+    Feature(name="ERC1155"),
+    Feature(name="SAFE_APPS"),
+    Feature(name="SPENDING_LIMIT"),
```
katspaugh, line 15: A couple suggestions, the same as for Wallets:
* `name` sounds a bit like something used for display. Maybe call it `key` or `feature_id`
* It might make sense to add an internal incremental id as the primary key instead of the name.

src/chains/models.py
fmrsabino: This was done on top of the branch that added the `Wallet`. Will rebase. The commit to be reviewed is 0a867b7. Since the base branch got merged now all the changes show up on this one.

src/chains/migrations/0032_feature.py
fmrsabino: Thanks for the suggestions. Made the changes here: 7ed430d

Uxio0 (approved):

jpalvarezl (commented):
I quickly looked at how django handles many to many relationships. I am just left with the question, what happens when you update a `Feature` db row, as we don't have id's I suppose a new entry is created in the db, right? Are we ok with this?

src/chains/serializers.py
```diff
@@ -80,6 +80,12 @@ class BlockExplorerUriTemplateSerializer(serializers.Serializer[Chain]):
     api = serializers.URLField(source="block_explorer_uri_api_template")
 
 
+class FeatureSerializer(serializers.Serializer[Feature]):
+    @staticmethod
+    def to_representation(instance: Feature) -> str:
```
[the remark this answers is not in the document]
fmrsabino: If you don't set the Primary Key in any of the fields then one is created for you (invisible to the model). You can see that an `id` is created automatically for you: https://github.com/gnosis/safe-config-service/pull/291/files#diff-01dbfb113f36cb35ffb2c95d2dcf8e3734f2f4e6d0909536465dba1d0f72afe4R37

From the model perspective, a `Feature` just has a single `key` (which represents the feature) and a single  `Feature` can be part of many chains

src/chains/models.py
```diff
@@ -132,3 +132,18 @@ class Wallet(models.Model):
...
+class Feature(models.Model):
+    # A feature can be enabled for multiple Chains and a Chain can have multiple features enabled
+    chains = models.ManyToManyField(
+        Chain, blank=True, help_text="Chains where this feature is enabled."
+    )
+    key = models.CharField(
+        unique=True,
+        max_length=255,
+        help_text="The unique name/key that identifies this feature",
+    )
```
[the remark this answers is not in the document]
fmrsabino: That's right. In order to maintain a `M2M` relationship a table is created. The migration performs the following:

```
CREATE TABLE "chains_feature" ("id" bigserial NOT NULL PRIMARY KEY, "key" varchar(255) NOT NULL UNIQUE);
CREATE TABLE "chains_feature_chains" ("id" bigserial NOT NULL PRIMARY KEY, "feature_id" bigint NOT NULL, "chain_id" bigint NOT NULL);
```

This Is invisible for us on the model level

fmrsabino:
> I quickly looked at how django handles many to many relationships. I am just left with the question, what happens when you update a Feature db row, as we don't have id's I suppose a new entry is created in the db, right? Are we ok with this?

`Feature` does have an `id` but it's hidden on the model side as it's mostly used for internal purposes. However this is not the case with `Wallet` since `Wallet` has the `key` as the primary key so changing it leads to a creation of a new entry. This is not what we want but a fix is on the way 👍

jpalvarezl (approved):
I just looked more closely at the migration and found the `id` field for `Feature` :) sorry for the self-induced confusion 😬

PR #293, opened 2021-11-05
Migrate Wallet table to one with BigAutoField

fmrsabino:
- Fix an issue where editing a `Wallet` would create a new one – the reason for this was because `name` was the primary key for `Wallet`. A primary key is ready-only so attempt to change would result in the creation of a new object.
- Change `Wallet.name` to `Wallet.key`
- Introducing an auto-generated ID to an already existing table can tricky as it expects default values for non-null columns (eg.: the new auto generated id). In order to tackle this the following migrations is performed:
  1. Create a new schema `WalletNew` with the columns `id`, `key` (old `name`) and `chains` (Many-to-Many relationship)
  2. Copy data from old schema `Wallet` and insert into the newly created schema `WalletNew`
  3. Clear relationship with between `Wallet <> Chains`
  4. Delete `Wallet` schema
  5. Rename `WalletNew` to `Wallet`


Uxio0 (approved):
I think this will make our lifes easier 😉

src/chains/migrations/0033_walletnew.py
```diff
@@ -0,0 +1,64 @@
...
+                    "id",
+                    models.BigAutoField(
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                (
+                    "key",
```
Uxio0, line 41: Why renaming to `key`? I found `name` more self explaining

fmrsabino: This was a suggestion from Ivan. The reason is that `name` might look like a UI specific field whereas `key` is more of an identifier. We can potentially add name to have a string representation of the `Wallet` in the future. What do you think?

fmrsabino:
@Uxio0 I added a test and fixed an issue with the migration (where it was using `name` to set the new fields instead of `key`) 26dce1b

src/chains/migrations/0033_walletnew.py
Uxio0: No strong opinion, but for me every wallet has a name, nothing related to the UI. Whatever you think is more clear 😃

Uxio0 (approved):

PR #299, opened 2021-11-08
Trigger CGW hook on Wallet and Feature update

fmrsabino:
- Trigger the CGW `/flush` endpoint when a `Feature` or `Wallet` is created, updated or deleted

### DIFF
--- a/src/chains/tests/test_signals.py
+++ b/src/chains/tests/test_signals.py
@@ -1,6 +1,7 @@
 import responses
 from django.test import TestCase, override_settings
 
+from chains.models import Feature, Wallet
 from chains.tests.factories import ChainFactory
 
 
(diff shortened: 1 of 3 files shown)


src/chains/tests/test_signals.py
```diff
@@ -91,3 +92,91 @@ def test_on_chain_update_with_no_flush_token_set(self) -> None:
...
+    @responses.activate
+    def test_on_feature_create_hook_call(self) -> None:
+        responses.add(
+            responses.POST,
+            "http://127.0.0.1/v1/flush/example-token",
+            status=200,
+            match=[responses.matchers.json_params_matcher({"invalidate": "Chains"})],  # type: ignore[no-untyped-call]
+        )
+
+        Feature(key="Test Feature").save()
```
fmrsabino, line 111: @Uxio0 the `@factory.post_generation` from `FeatureFactory` seems to create and update the feature in one go (probably due to setting the M2M relationship after creation).

So `FeatureFactory.create()` results in two signals while `Feature(key="Test Feature").save()` results in one (as expected).

Uxio0: Yes, I've seen that before 😉 . One thing to take into consideration is that `bulk` requests don't trigger signals

Uxio0 (approved):

PR #317, opened 2021-11-17
Integrate mypy in safe-apps app

fmrsabino:
- Add types to `safe-apps` app
- Check `safe-apps` using `--strict` mode in Github Actions


Uxio0 (approved):
Good job

PR #362, opened 2021-12-15
Added the issue templates we have in the CGW

jpalvarezl:
These are the same issue templates currently being used in the [CGW](https://github.com/gnosis/safe-client-gateway/tree/main/.github/ISSUE_TEMPLATE)


Uxio0 (approved):
Can I get that on the tx service?

jpalvarezl:
@Uxio0 I have opened a PR already :) https://github.com/gnosis/safe-transaction-service/pull/560

PR #363, opened 2021-12-16
Feature: Safe App access policy fields

### LINKED ISSUE #351: Safe apps: add a field for different access control types
Context: https://github.com/gnosis/safe-react/issues/2672

We need to know if the app restricts access by source (and therefore can't be loaded in the desktop app) to filter such apps.

The field should be an array of accepted origins and NULL if the app doesn't restrict its access.

The field should be
```
access_control: { type: "NO_RESTRICTIONS" }
or
access_control: { type: 'DOMAIN_RESTRICTION', data: [... ] }
```

It should also accept a new query param `host_domain` that accepts an URL and returns apps that can be displayed for this host

- [ ] Implementation
- [ ] Gitbook doc flag update

mikheevm:
This pull request adds a new Client table (only the `url` field for now) and a field `exclusive_clients` to the Safe App model.

It also adds filtering to the view to allow clients to specify the domain URL in the `GET` request and receive apps guaranteed to work on the client on this domain.

Closes #351 

Original issue https://github.com/gnosis/safe-react/issues/2672

Example request/response:
```
GET /api/v1/safe-apps/?client_url=safe.com

[
   {
      "id":0,
      "url":"https://shaw.net/",
      "name":"Reid, Sanchez and Bailey",
      "iconUrl":"https://placekitten.com/71/525",
      "description":"Focused fresh-thinking Graphic Interface",
      "chainIds":[
         956
      ],
      "provider":"None",
      "accessControl":{
         "type":"DOMAIN_ALLOWLIST",
         "value":[
            "safe.com"
         ]
      }
   },
   {
      "id":1,
      "url":"http://www.hernandez-gonzalez.com/",
      "name":"Haynes LLC",
      "iconUrl":"https://dummyimage.com/71x402",
      "description":"Front-line foreground open architecture",
      "chainIds":[
         5737,
         9036
      ],
      "provider":"None",
      "accessControl":{
         "type":"NO_RESTRICTIONS"
      }
   }
]
```


src/safe_apps/tests/test_views.py
```diff
@@ -178,6 +219,126 @@ def test_apps_returned_on_same_key_pair(self) -> None:
...
+
+        response = self.client.get(path=url, data=None, format="json")
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), json_response)
+
+    def test_apps_returned_on_filtered_client_url(self) -> None:
+        client = ClientFactory.create(url="safe.com")
+        client_2 = ClientFactory.create(url="pump.com")
+        safe_app_1 = SafeAppFactory.create(exclusive_clients=(client,))
```
rmeissner, line 303: Should we add a test entry that has both clients set and should then also be returned?

src/safe_apps/serializers.py
```diff
@@ -23,4 +37,10 @@ class Meta:
             "description",
             "chain_ids",
             "provider",
+            "access_control",
         ]
+
+    @staticmethod
+    @swagger_serializer_method(serializer_or_field=AccessControlPolicySerializer)  # type: ignore[misc]
+    def get_access_control(obj: SafeApp) -> ReturnDict:
+        return AccessControlPolicySerializer(obj).data
```
fmrsabino, line 46: On the API level, we are always returning `value` even though we know it's going to be empty if `NO_RESTRICTIONS`. If we always end up exposing both `type` and `value` the clients might want to check both even though `NO_RESTRICTIONS` should always have `"value" : []`.

Suggestion: introduce two serialisers (one for each type). You'd check the type on this function and serialize accordingly:

- if `NO_RESTRICTIONS`

```javascript
"accessControl" : {
  "type": "NO_RESTRICTIONS"
}
```

- if `DOMAIN_ALLOWLIST`

```javascript
"accessControl" : {
  "type": "DOMAIN_ALLOWLIST",
  "value" : [ ... ] // domains
}
```
On the client side, they can then associate each type with a different payload. Eg.: if the client gets `DOMAIN_ALLOWLIST` then it should map to a type that has a collection of domains.

src/safe_apps/tests/factories.py
```diff
@@ -24,3 +31,14 @@ class Meta:
...
     provider = None
+
+    @factory.post_generation
+    def exclusive_clients(self, create, extracted, **kwargs):  # type: ignore[no-untyped-def] # decorator is untyped
+        if not create:
+            # Simple build, do nothing.
+            return
+
+        if extracted:
+            # A list of apps was passed in, use them
```
fmrsabino, line 42:
```suggestion
            # A list of clients was passed in, use them
```

src/safe_apps/tests/test_views.py
```diff
@@ -29,6 +30,10 @@ def test_json_payload_format(self) -> None:
                 "description": safe_app.description,
                 "chainIds": safe_app.chain_ids,
                 "provider": None,
+                "accessControl": {
+                    "type": SafeApp.AccessControlPolicy.NO_RESTRICTIONS,
```
fmrsabino, line 34: Instead of using `SafeApp.AccessControlPolicy.NO_RESTRICTIONS`, use the expected `str` on the payload side. This way, we can catch if the API was broken due to a refactor (eg.: if we rename `SafeApp.AccessControlPolicy.NO_RESTRICTIONS` everywhere then the test would be successful even though it might not be the result that we wanted)

src/safe_apps/tests/test_views.py
```diff
@@ -178,6 +219,126 @@ def test_apps_returned_on_same_key_pair(self) -> None:
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.json(), json_response)
 
+    def test_apps_returned_on_non_existent_client_url(self) -> None:
+        safe_app = SafeAppFactory.create()
+        url = reverse("v1:safe-apps:list") + f'{"?client_url=non_existent_host"}'
```
fmrsabino, line 224: Why the query parameter here? We don't do any custom handling with it right?

fmrsabino (changes requested):

src/safe_apps/tests/test_views.py
mikheevm: It's there to check how the API reacts if you pass an unknown client URL. The `client_url` param is used to return all the apps that work on passed URL

fmrsabino: Sorry you are right. I see that you added it on the view side 👍

src/safe_apps/models.py
```diff
@@ -10,7 +10,21 @@ def __str__(self) -> str:
         return f"{self.name} | {self.url}"
 
 
+class Client(models.Model):
+    url = models.URLField(
+        unique=True,
+        help_text="The domain URL client is hosted at",
+    )
```
fmrsabino, line 17: Wondering if `models.URLField` is the best field to be used for the constraints that we have on the query side.

would for example: `http://example.com` be different than `https://example.com`.
Or is it relevant that `https://example.com/path/1` is not the same as `https://example.com`

The restrictions set on the access policy seem to be more on the domain level correct? Going with just a pure domain might also be more clear on the query parameter side: `?client_url=http://example.com` vs `?client_url=https://example.com`?

src/safe_apps/tests/test_views.py
mikheevm: I adjusted the existing test to include such app as well

src/safe_apps/models.py
mikheevm: This makes sense to me. Is there a pre-defined domain field for django models? Or I should just go with CharField + custom validator?

mikheevm: https://code.djangoproject.com/ticket/18120 `wontfix` :/

fmrsabino: I would say a `CharField` with a custom validator should be enough. If we see that we use the same pattern in more places we can think on creating a custom `Field`.

src/safe_apps/models.py
```diff
@@ -1,7 +1,54 @@
...
+    host_re = "(" + hostname_re + domain_re + tld_re + ")"
+    regex = (
+        r"(?:" + ipv4_re + "|" + ipv6_re + "|" + host_re + ")"
+        r"(?::\d{2,5})?"  # port
+        r"\Z"
+    )
+    return RegexValidator(
+        regex,
+        message="Enter a valid hostname (Without a resource path)",
```
rmeissner, line 44: Also without a protocol?

mikheevm: the protocol is considered valid now

src/safe_apps/models.py
```diff
@@ -1,7 +1,54 @@
...
+
 from django.contrib.postgres.fields import ArrayField
+from django.core.validators import RegexValidator
 from django.db import models
 
 
+def _hostname_validator() -> RegexValidator:
+    """
+    Proudly taken from https://github.com/MehdioKhan/django-domainname
+    """
```
fmrsabino, line 11: This is licensed under GPL so it cannot be redistributed under an MIT license.

Isn't it possible to have simpler regex for this? I see that port numbers are also covered. Do we need those?

mikheevm: > This is licensed under GPL so it cannot be redistributed under an MIT license.

Hmm, the code itself is a slightly modified URLValidator from Django with dropped additional checks 🤷‍♂️ 

> Isn't it possible to have simpler regex for this? I see that port numbers are also covered. Do we need those?

I believe it is possible, but from studying existing validators, I understood that it's not as simple as it seems, and I decided not to reinvent the wheel because it's guaranteed I'll miss some edge-case. Anyway, I've just pushed a simpler regex that seems to work for our problem.

src/safe_apps/models.py
```diff
@@ -1,6 +1,17 @@
+from enum import Enum
+
 from django.contrib.postgres.fields import ArrayField
+from django.core.validators import RegexValidator
 from django.db import models
 
+HostnameValidator = RegexValidator(
```
fmrsabino, line 7: Since this is a constant we should follow PEP8 conventions for the naming https://www.python.org/dev/peps/pep-0008/#constants

Suggestion: `_HOSTNAME_VALIDATOR`

src/safe_apps/migrations/0007_safeapp_access_policy.py
```diff
@@ -0,0 +1,52 @@
...
+                ),
+                (
+                    "url",
+                    models.CharField(
+                        help_text="The domain URL client is hosted at",
+                        max_length=255,
+                        unique=True,
+                        validators=[
+                            django.core.validators.RegexValidator(
+                                "^(https?:\\/\\/)?(www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{2,256}\\.[a-z]{2,6}\\/?$",
```
fmrsabino, line 34: This regex is different than the one set in the model 🤔 

Maybe it's better to recreate this migration:

```shell
python src/manage.py migrate safe_apps 0006
rm src/safe_apps/migrations/0007_safeapp_access_policy.py
python src/manage.py makemigrations
```

fmrsabino (changes requested):
I think the new regex looks much better! 🚀 

Left some questions/suggestions

src/safe_apps/models.py
mikheevm: I'm curious how I got through that Linter Castle...

src/safe_apps/migrations/0007_safeapp_access_policy.py
mikheevm: It's the same, isn't it? I recreated the migration before pushing and did it once again but the output is the same

mikheevm: https://github.com/gnosis/safe-config-service/pull/363/commits/3be622b83cd22b5f90557229b34f888774426e11

fmrsabino (commented):
Looks good 👏 

Just one question. Did you try adding a `Client` via the admin interface (does it work as expected with the many-to-many relationship)?

mikheevm:
@fmrsabino good catch, I actually forgot to register it to the admin panel. I fixed and it and tested it
Screenshot 2021-12-31 at 17 19 49
 - works as expected
Screenshot 2021-12-31 at 17 19 42

fmrsabino (approved):
🚀 🚀 🚀

mikheevm:
Thanks for the thorough review, @fmrsabino. Is it good to go now, or should we wait for a review from Uxio?

fmrsabino:
> Thanks for the thorough review, @fmrsabino. Is it good to go now, or should we wait for a review from Uxio?

Let's merge it 🙌 We can start testing with the clients on staging!

PR #371, opened 2021-12-22
Set OTP for admin access

### LINKED ISSUE #341: Enforce 2FA for the admin interface
We should enforce 2fa for all accounts that have access to the admin interface.

fmrsabino:
Closes #341 

- Enable OTP for admin access – this setting can be enabled via `DJANGO_OTP_ADMIN` (default `true`)
- If `DJANGO_OTP_ADMIN` is set to `true` then a new admin (`OTPAdminConfig`) is plugged in and logging in requires an OTP token
- If `DJANGO_OTP_ADMIN` is set to `false` then the default Django admin is used – this is useful when developing locally for example
- When deploying the OTP admin, current admins will be "locked out" of their account because an OTP will be required to login, to get around this issue a static OTP can be configured for one of the admins to do the initial setup via `python src/manage.py addstatictoken <username>`. Once every admin has the OTP set then the static token can be removed

### DIFF
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -49,7 +49,6 @@ INSTALLED_APPS = [
     "about.apps.AboutAppConfig",
     "chains.apps.AppsConfig",
     "safe_apps.apps.AppsConfig",
-    "django.contrib.admin",
     "django.contrib.auth",
     "django.contrib.contenttypes",
     "django.contrib.sessions",
@@ -57,6 +56,9 @@ INSTALLED_APPS = [
     "django.contrib.staticfiles",
     "rest_framework",
     "drf_yasg",
+    "django_otp",
+    "django_otp.plugins.otp_totp",
+    "django_otp.plugins.otp_static",
 ]
 
 MIDDLEWARE = [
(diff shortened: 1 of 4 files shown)


src/config/settings.py
```diff
@@ -67,10 +69,19 @@
...
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
 ]
 
+DJANGO_OTP_ADMIN = strtobool(os.getenv("DJANGO_OTP_ADMIN", "true"))
+if DJANGO_OTP_ADMIN:
+    # Use OTP admin
+    INSTALLED_APPS.append("config.admin.OTPAdminConfig")
+else:
+    # Use Default admin
+    INSTALLED_APPS.append("django.contrib.admin")
```
fmrsabino, line 83: @Uxio0 Wondering what is the best way to test this. Overriding the settings in the tests won't work since settings are loaded when the Django initialises.

It looks like we need a test that loads Django after setting the environment variable `DJANGO_OTP_ADMIN`. And then we can check the installed applications.

Uxio0: You can enable it always for tests, does it breaks your admin tests?

fmrsabino: The tests don't break but the coverage decreased because we are not testing this `if/else` branch 😅 

I think for this it would require a test where we set an environment variable and programatically launch Django. This can be probably be tackled later if required.

Uxio0: If the only problem is coverage, you can use: https://coverage.readthedocs.io/en/coverage-4.3.3/excluding.html#excluding-code-from-coverage-py I think it's for cases like this one. I think trying to test this can give us more trouble than the problem we are trying to fix

Uxio0 (approved):

PR #374, opened 2021-12-24
Restrict native currency logo dimensions

### LINKED ISSUE #373: Token logos are too large/heavy
**Describe the bug**
I did a spot check on different coin icons we are receiving from the gateway.

https://safe-transaction-assets.gnosis-safe.io/chains/4/currency_logo.png – 3258x3258 px, 161 Kb
https://safe-transaction-assets.gnosis-safe.io/chains/43114/currency_logo.png – 1500x1500, 41 Kb

I suppose many other icons are also unnecessarily big. This is very bad for the page loading speed.

## Frontend requirements
* Token logos should be sufficiently big, so that we can downscale them to different sizes in different components.
* But not too big. 512x512 would be good enough for most situations, including hi-res (aka Retina) screens.
* Images should be optimized, e.g. via OptiPng. This will already halve the current sizes.
 
Try it on optipng.com and there's a CLI util of the same name.

## To summarize
Please do the following:
* Resize all the existing icons to 512x512 px
* If the icon is smaller, keep the current size
* Run them through optipng
* Set up a pipeline to do the same for new icons

## Automated pipeline
Whenever we upload an image, we should programmatically perform the same actions as requested above for existing images.
Namely:
* resize it to 512 px on the largest side (e.g. 512x512, or 300x512, or 512x400 etc)
* if the original is smaller, keep the original size
* optimize using OptiPng

## Why OptiPng
OptiPng is a de-facto standard CLI tool to remove unnecessary kilobytes from PNGs and JPGs.
In most cases, it reduces the size of a file by 100% (i.e. halves the size).
The resource cost is negligible as this will be happening on the upload of a single image.

fmrsabino:
Closes #373

- `currency_logo_uri` size is now validated – if its image `width` or `height` is greater than `512` then a `ValidationError` is thrown


src/chains/models.py
```diff
@@ -24,6 +26,12 @@ def native_currency_path(instance: "Chain", filename: str) -> str:
     return f"chains/{instance.id}/currency_logo{file_extension}"
 
 
+def validate_native_currency_size(image: str | IO[bytes]) -> None:
+    image_width, image_height = get_image_dimensions(image)
+    if image_width > 512 or image_height > 512:
```
jpalvarezl, line 31: Should we restrict the ratio too? meaning `and image_width == image_height` ? Seems to me that this is expected by the clients, but I don't know it for a fact. Pointing it out just in case :)

jpalvarezl (approved):

src/chains/models.py
fmrsabino: Yeah that's a good point. I will review the images and check if we have any with a different ratio

PR #384, opened 2022-01-03
Add GET /safe-apps client_url param to swagger, align chainId param/variable naming

mikheevm:
This PR fixes:
- In my previous PR for access control for Safe Apps I forgot to include client_url param to swagger parameters
- (nit) Aligns naming between chainId param and the variable naming
- convert `client_url` to camelCase in tests/views

### DIFF
--- a/src/safe_apps/tests/test_views.py
+++ b/src/safe_apps/tests/test_views.py
@@ -210,7 +210,7 @@ class FilterSafeAppListViewTests(APITestCase):
 
     def test_apps_returned_on_non_existent_client_url(self) -> None:
         safe_app = SafeAppFactory.create()
-        url = reverse("v1:safe-apps:list") + f'{"?client_url=non_existent_host"}'
+        url = reverse("v1:safe-apps:list") + f'{"?clientUrl=non_existent_host"}'
 
         response = self.client.get(path=url, data=None, format="json")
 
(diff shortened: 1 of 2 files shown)


mikheevm:
@fmrsabino I just realized that `chainId` and `client_url` use different naming conventions. Should I also align them to use camelCase or something else?

fmrsabino:
> @fmrsabino I just realized that `chainId` and `client_url` use different naming conventions. Should I also align them to use camelCase or something else?
> 
> EDIT: Ok, it looks like we use camelCase in responses, so it makes sense to stick to camelCase in params as well. I just pushed a commit to align `client_url`

Yes 👍 Even though we use `_` in the python parameters, it goes through a camel case middleware which Is installed then converts it to `camelCase` for the clients

fmrsabino (approved):

mikheevm:
> Yes 👍 Even though we use _ in the python parameters, it goes through a camel case middleware which Is installed then converts it to camelCase for the clients

It is only true for responses, isn't it?

fmrsabino:
> > Yes 👍 Even though we use _ in the python parameters, it goes through a camel case middleware which Is installed then converts it to camelCase for the clients
> 
> It is only true for responses, isn't it?

Yes. For the query parameters we do it as you just did in this PR

PR #387, opened 2022-01-06
Bug: apps are filtered by the client if clientUrl param is empty

mikheevm:
When writing tests for the safe-client-gateway, I discovered that if you specify an empty `clientUrl` query param, the apps would still get filtered by an empty string. In this PR:
- add a check for an empty `clientUrl` param and extend the test
- rename the variable for `chainId` param from `network_id` to `chain_id` to match the param name and use the same approach for client_url
- Use `Union` for type definitions instead of `|` to remove hard requirements on python 3.10+

### DIFF
--- a/src/safe_apps/views.py
+++ b/src/safe_apps/views.py
@@ -47,7 +47,7 @@ class SafeAppsListView(ListAPIView):
             queryset = queryset.filter(chain_ids__contains=[chain_id])
 
         client_url = self.request.query_params.get("clientUrl")
-        if client_url is not None and client_url:
+        if client_url:
             queryset = queryset.filter(
                 Q(exclusive_clients__url=client_url) | Q(exclusive_clients__isnull=True)
             )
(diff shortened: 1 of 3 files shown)


src/safe_apps/views.py
```diff
@@ -42,14 +42,14 @@ def get(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
-        if network_id is not None and network_id.isdigit():
-            queryset = queryset.filter(chain_ids__contains=[network_id])
+        chain_id = self.request.query_params.get("chainId")
+        if chain_id is not None and chain_id.isdigit():
+            queryset = queryset.filter(chain_ids__contains=[chain_id])
 
-        host = self.request.query_params.get("clientUrl")
-        if host is not None:
+        client_url = self.request.query_params.get("clientUrl")
+        if client_url is not None and client_url:
```
fmrsabino, line 50: Wouldn't it be enough to have 

```python
if client_url:
```

Probably the same can be done above too:

```python
if chain_id.isdigit():
```

mikheevm: for the `client_url` it works, thanks for the tip.

I think you cannot call `isdigit` method on `None`, but I can specify an empty string as the default value in param getter, wdyt?

fmrsabino: I'd say to leave both conditions in (ie.: not have the default for now)

fmrsabino (approved):
🚀

PR #395, opened 2022-01-17
Change chains pagination max_limit to 20

fmrsabino:
- Change Chains pagination limit from `100` to `20`

### DIFF
--- a/src/chains/views.py
+++ b/src/chains/views.py
@@ -14,7 +14,7 @@ from .serializers import ChainSerializer
 class ChainsListView(ListAPIView):
     serializer_class = ChainSerializer
     pagination_class = LimitOffsetPagination
-    pagination_class.max_limit = 100
+    pagination_class.max_limit = 20
     pagination_class.default_limit = 20
     queryset = Chain.objects.all()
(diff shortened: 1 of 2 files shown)


src/chains/views.py
```diff
@@ -14,7 +14,7 @@
 class ChainsListView(ListAPIView):
     serializer_class = ChainSerializer
     pagination_class = LimitOffsetPagination
-    pagination_class.max_limit = 100
+    pagination_class.max_limit = 20
```
Uxio0, line 17: Isn't this changing the class attributes for every view using it?

Uxio0: So I guess every view that has the pagination class instantiated after this will have `max_limit=20`

fmrsabino: Yes you are right. I think it's better to remove this for now and add it to the Django configuration side 🤦

Uxio0: I wouldn't do that there. I would define a custom LimitOffsetPagination and set the values there

fmrsabino: Done a72f7f1

Uxio0 (approved):

PR #401, opened 2022-01-18
Fix safe-apps view tests (sorting)

fmrsabino:
- Currently there's no sorting in place for the safe-apps endpoints so there's no guarantee on how the different safe-apps will be returned
- Use `self.assertCountEqual` to compare the entries present in the response payload regardless of their order


mikheevm (approved):
💪

PR #402, opened 2022-01-19
support for trusted csrf origins

InoMurko:
Based on https://docs.djangoproject.com/en/4.0/ref/settings/#csrf-trusted-origins I could use this setting and I don't think it harms being here.

### DIFF
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -230,5 +230,8 @@ DEFAULT_FILE_STORAGE = os.getenv(
 
 # SECURITY
 # https://docs.djangoproject.com/en/4.0/ref/settings/#csrf-trusted-origins
-allowed_csrf_origins = os.getenv("CSRF_TRUSTED_ORIGINS", "")
-CSRF_TRUSTED_ORIGINS = [allowed_csrf_origins.strip() for allowed_csrf_origins in allowed_csrf_origins.split(",")]
+allowed_csrf_origins = os.getenv("CSRF_TRUSTED_ORIGINS", "http://127.0.0.1")
+CSRF_TRUSTED_ORIGINS = [
+    allowed_csrf_origins.strip()
+    for allowed_csrf_origins in allowed_csrf_origins.split(",")
+]
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -230,8 +230,9 @@ DEFAULT_FILE_STORAGE = os.getenv(
 
 # SECURITY
 # https://docs.djangoproject.com/en/4.0/ref/settings/#csrf-trusted-origins
-allowed_csrf_origins = os.getenv("CSRF_TRUSTED_ORIGINS", "http://127.0.0.1")
-CSRF_TRUSTED_ORIGINS = [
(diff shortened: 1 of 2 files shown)


fmrsabino (commented):
Looks good.

Can we add an entry to the `.env.sample` too? We have that file also to document the environment setup.

Something similar to https://github.com/gnosis/safe-config-service/blob/704d478875d2147beacf74c93916a5f98db462c8/.env.example#L29-L36

But with the value commented:
```
#CSRF_TRUSTED_ORIGINS=""
```

This way it can be quickly setup by copying the `.env.sample` to `.env` and uncommenting this value if needed.

fmrsabino:
Additionally, I'm configuring the actions with the CLA so that you can merge the PR after approval. I'll ping it here when it's ready. You might need to agree to the same CLA that you did for the other repos.

InoMurko:

fmrsabino (dismissed):
Awesome! 🚀 

Can you rebase against the latest `main` branch. We just pushed the following https://github.com/gnosis/safe-config-service/commit/244d4b34b9c3eef30f90e55eba346a8c67939ddd (which would enable the actions to be run on your PR)

fmrsabino (approved):

src/config/settings.py
```diff
@@ -227,3 +227,11 @@
 DEFAULT_FILE_STORAGE = os.getenv(
     "DEFAULT_FILE_STORAGE", "storages.backends.s3boto3.S3Boto3Storage"
 )
+
+# SECURITY
+# https://docs.djangoproject.com/en/4.0/ref/settings/#csrf-trusted-origins
+allowed_csrf_origins = os.getenv("CSRF_TRUSTED_ORIGINS", "http://127.0.0.1")
```
fmrsabino, line 233: It seems that Django 4 introduced new checks to this value: https://docs.djangoproject.com/en/4.0/releases/4.0/#csrf-trusted-origins-changes-4-0

So `CSRF_TRUSTED_ORIGINS` cannot be empty. Either it is set with the correct constraints or not set at all.

I would probably wrap the setting of `CSRF_TRUSTED_ORIGINS` under the condition that it has a non-null/non-blank value.

ie.: I'd keep the Django defaults as much as possible.

fmrsabino (changes requested):

src/config/settings.py
InoMurko: Which is funny, because the default is in fact an empty list https://docs.djangoproject.com/en/4.0/ref/settings/#csrf-trusted-origins

fmrsabino: True, which with your changes should've been fine since `CSRF_TRUSTED_ORIGINS` would just be empty. But it seems that there's a check if it is set by the application and if It is then it cannot be empty which is strange.

```
?: (4_0.E001) As of Django 4.0, the values in the CSRF_TRUSTED_ORIGINS setting must start with a scheme (usually http:// or https://) but found . See the release notes for details.
```

fmrsabino (approved):
🚀

PR #410, opened 2022-01-26
Update CODEOWNERS

fmrsabino:
- Set @gnosis/safe-services as a Code Owner for the Safe Config Service
- Set @mikheevm as a Code Owner for the `safe-apps` Django app


mikheevm (approved):
☺️☺️☺️🙌

PR #451, opened 2022-03-10
Autodeploy auth

gMonty030:
- Removed Dockerhub autodeploy weebhooks
- Added new step for the autodeploy request after the image is built
- Added auth to autodeploy service, now is more secure the deployment of new versions.
- Added nginx auth script

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -170,7 +170,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Deploy Staging
-      if: github.ref == 'refs/heads/main'
       run: bash scripts/autodeploy.sh
       env:
         AUTODEPLOY_URL: ${{ secrets.AUTODEPLOY_URL }}


Uxio0 (dismissed):

gMonty030:
- - -
   ***I have read the CLA Document and I hereby sign the CLA***
   - - -

.github/workflows/test.yml
```diff
@@ -162,3 +162,24 @@ jobs:
...
+        AUTODEPLOY_URL: ${{ secrets.AUTODEPLOY_URL }}
+        AUTODEPLOY_TOKEN: ${{ secrets.AUTODEPLOY_TOKEN }}
+        TARGET_BRANCH: "staging"
+    - name: Deploy Develop
+      if: github.ref == 'refs/heads/develop'
+      run: bash scripts/autodeploy.sh
+      env:
+        AUTODEPLOY_URL: ${{ secrets.AUTODEPLOY_URL }}
+        AUTODEPLOY_TOKEN: ${{ secrets.AUTODEPLOY_TOKEN }}
+        TARGET_BRANCH: "develop"
```
fmrsabino, line 185: There is no `develop` branch.

.github/workflows/test.yml
```diff
@@ -162,3 +162,24 @@ jobs:
...
+    if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/tags/')
+    steps:
+    - uses: actions/checkout@v2
+    - name: Deploy Staging
+      if: github.ref == 'refs/heads/main'
+      run: bash scripts/autodeploy.sh
+      env:
+        AUTODEPLOY_URL: ${{ secrets.AUTODEPLOY_URL }}
+        AUTODEPLOY_TOKEN: ${{ secrets.AUTODEPLOY_TOKEN }}
+        TARGET_BRANCH: "staging"
```
fmrsabino, line 178: Wouldn't `IMAGE_TAG` or similar be more representative of this field? (we don't have a branch called `staging`)

.github/workflows/test.yml
```diff
@@ -162,3 +162,24 @@ jobs:
           mv /tmp/.buildx-cache-new /tmp/.buildx-cache
       - name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
+
+  autodeploy:
+    runs-on: ubuntu-latest
+    needs: [docker-deploy]
```
fmrsabino, line 168: `docker-deploy` is not a declared job or did I miss something?

fmrsabino (changes requested):
I'm still missing the background context for this PR.

I see a `curl` but most of the data is behind environment variables so it's harder to get the context from the script.

.github/workflows/test.yml
gMonty030: Yeah, it was in the case you are going to deploy develop in the future. At least the other projects that I checked they have it. I can remove it even so.

.github/workflows/test.yml
gMonty030: I'll change it, this comes from other autodeploy projects I didn't want to change the var just because Devops team may be familiarized but completely agree, another could be better.

.github/workflows/test.yml
gMonty030: About the background context, this is a change for the auto-deploy process, you had Dockerhub as the trigger but it didn't have auth method available, so because of security concerns we decide to use the Nginx auth of the auto-deploy server.

.github/workflows/test.yml
gMonty030: it's the name of the step before this, it has the same name as the rest of the projects.

fmrsabino: The name in this project is `docker-publish`

.github/workflows/test.yml
fmrsabino: > About the background context, this is a change for the auto-deploy process, you had Dockerhub as the trigger but it didn't have auth method available, so because of security concerns we decide to use the Nginx auth of the auto-deploy server.

The current Docker push action available: `docker/login-action@v1.13.0` has a combination of username and password. But an auth token can also be used instead of the password.

Wouldn't that be the same security-wise to the url+token combination?

fmrsabino: > Yeah, it was in the case you are going to deploy develop in the future. At least the other projects that I checked they have it. I can remove it even so.

I would remove it. Having a `develop` branch is a bigger change in our workflow so if we ever add it we'd also need to review the pipeline steps required.

gMonty030:

gMonty030:

.github/workflows/test.yml
gMonty030: True, my fault, I corrected it. Thanks

.github/workflows/test.yml
gMonty030: > I would remove it. Having a develop branch is a bigger change in our workflow so if we ever add it we'd also need to review the pipeline steps required.

I removed the develop branch

gMonty030: > The current Docker push action available: docker/login-action@v1.13.0 has a combination of username and password. But an auth token can also be used instead of the password.

Wouldn't that be the same security-wise to the url+token combination?

.github/workflows/test.yml
```diff
@@ -162,3 +162,17 @@ jobs:
           mv /tmp/.buildx-cache-new /tmp/.buildx-cache
       - name: Image digest
         run: echo ${{ steps.docker_build.outputs.digest }}
+
+  autodeploy:
+    runs-on: ubuntu-latest
+    needs: [docker-publish]
+    if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/tags/')
```
fmrsabino, line 169: This job has a step where it only executes if `github.ref == 'refs/heads/main'` so this job would be triggered for other events that would not execute any steps.

I would do the following:

```suggestion
    github.event_name == 'push' && github.ref == 'refs/heads/main'
```

and remove the `if` condition from the `Deploy Staging` step.

fmrsabino (changes requested):

.github/workflows/test.yml
gMonty030: The main goal was to avoid the calls to the autodeploy server without authentication, now the requests (script file) has those credentials and make more secure this step

.github/workflows/test.yml
```diff
@@ -162,3 +162,17 @@ jobs:
...
         run: echo ${{ steps.docker_build.outputs.digest }}
+
+  autodeploy:
+    runs-on: ubuntu-latest
+    needs: [docker-publish]
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    steps:
+    - uses: actions/checkout@v2
+    - name: Deploy Staging
+      if: github.ref == 'refs/heads/main'
```
fmrsabino, line 173: As mentioned in https://github.com/gnosis/safe-config-service/pull/451#discussion_r827891120

The following can be removed:

```suggestion
```

fmrsabino (changes requested):

fmrsabino (approved):

PR #468, opened 2022-03-23
Add support for non-TLD urls

### LINKED ISSUE #231: Config service does not accept urls without TLD
- Docker maps hostnames to the ip addresses, so a valid url could be http://redis
- Only exception is `localhost`
- https://stackoverflow.com/questions/49102410/django-urlfield-doesnt-accept-hostname-only-urls

fmrsabino:
Closes #231

- Add support for URLs without TLD in `transaction_service_uri` and `vpc_transaction_service_uri`
- The protocol is restricted to `http` and `https`


src/chains/models.py
```diff
@@ -32,6 +33,21 @@ def validate_native_currency_size(image: Union[str, IO[bytes]]) -> None:
         raise ValidationError("Image width and height need to be at most 512 pixels")
 
 
+def validate_tx_service_url(url: str) -> None:
+    result = urlparse(url)
+    if not all(
+        [
```
Uxio0, line 39: Small tip: I think you can remove the `[]` comprehension here and use a generator `()`. That way if the first evaluates to `False` it will not continue. In this case is not relevant, though, as it's a really fast check

src/chains/tests/test_models.py
```diff
@@ -159,6 +159,41 @@ def test_valid_addresses(self) -> None:
...
+
+            with self.subTest(msg=f"{invalid_url} is not a valid url"):
+                with self.assertRaises(ValidationError):
+                    chain = ChainFactory.create(vpc_transaction_service_uri=invalid_url)
+                    chain.full_clean()
+
+    def test_valid_urls(self) -> None:
+        param_list = [
+            "http://tx-service",
+            "https://tx-service",
```
Uxio0, line 183: Maybe you can add one with port `https://tx-service:8000`

Uxio0 (changes requested):

src/chains/models.py
fmrsabino: Done in [d381e3b](https://github.com/gnosis/safe-config-service/pull/468/commits/d381e3ba966f02b54f8766ca862673c356f8cc20)

src/chains/tests/test_models.py
fmrsabino: Done in [d381e3b](https://github.com/gnosis/safe-config-service/pull/468/commits/d381e3ba966f02b54f8766ca862673c356f8cc20)

Uxio0 (approved):

Uxio0:
Can we do a release so I can fix this issue on the safe-infra repo?

PR #469, opened 2022-03-23
Extract Client Gateway cache invalidation

fmrsabino:
- Extracts Safe Client Gateway related logic to `src/clients/safeclientgateway.py`
- This allows for other Django applications to use this client


src/chains/signals.py
```diff
@@ -1,72 +1,48 @@
...
-        post = setup_session().post(
-            url,
-            json={"invalidate": "Chains"},
-            headers={"Authorization": f"Basic {cgw_flush_token}"},
-        )
-        post.raise_for_status()
-    except Exception as error:
-        logger.error(error)
+def _flush_cgw_chains() -> None:
+    clients.safeclientgateway.flush(
```
Uxio0, line 16: camel case?

src/chains/signals.py
```diff
@@ -1,72 +1,48 @@
...
+        json={"invalidate": "Chains"},
+    )
 
 
 @receiver(post_save, sender=Chain)
 @receiver(post_delete, sender=Chain)
 def on_chain_update(sender: Chain, **kwargs: Any) -> None:
     logger.info("Chain update. Triggering CGW webhook")
-    _trigger_client_gateway_flush()
+    _flush_cgw_chains()
```
Uxio0, line 27: Why not using the same method with multiple receivers?

src/clients/safeclientgateway.py
```diff
@@ -0,0 +1,39 @@
+import logging
```
Uxio0, line 1: Filename must be camel case

Uxio0 (changes requested):

src/chains/signals.py
fmrsabino: Shouldn't package/modules be all-lowercase? https://peps.python.org/pep-0008/#package-and-module-names

src/clients/safeclientgateway.py
fmrsabino: > Filename must be camel case

Same as here: https://github.com/gnosis/safe-config-service/pull/469#discussion_r833149938

src/chains/signals.py
Uxio0: You can do that depending on the `sender`, but I get it

src/chains/signals.py
Uxio0: Sorry. I was talking about snake case 😅

fmrsabino: Ah this is because it's not intended to be used outside of `signals.py`.

I looked at this paragraph:

> “Private” instance variables that cannot be accessed except from inside an object don’t exist in Python. However, there is a convention that is followed by most Python code: a name prefixed with an underscore (e.g. _spam) should be treated as a non-public part of the API (whether it is a function, a method or a data member). It should be considered an implementation detail and subject to change without notice.

https://docs.python.org/3.10/tutorial/classes.html#private-variables

Uxio0: I'm taking about `safeclientgateway` 😅

fmrsabino: [dfda6b1](https://github.com/gnosis/safe-config-service/pull/469/commits/dfda6b1f9e3fb6f4d75390c81c92a17c9c3043cd)

src/clients/safeclientgateway.py
fmrsabino: It doesn't need to be but I changed it [dfda6b1](https://github.com/gnosis/safe-config-service/pull/469/commits/dfda6b1f9e3fb6f4d75390c81c92a17c9c3043cd)

Uxio0 (approved):

PR #496, opened 2022-04-06
Add tags to SafeApp

### LINKED ISSUE #486: Add tag support to Safe Apps
Tags are being introduced in order to provide more context for each Safe App.

Currently all Safe Apps are part of the same "category" ie.: there's no way to group or label Safe Apps.

Labelling Safe Apps has the following advantages:
- Provides filtering capabilities on a specific label (eg.: DeFi apps)
- Can be used to promote some apps (eg.: a promoted group)
- Can be used to distinguish official apps from not official ones

## API
```javascript
GET /api/v1/safe-apps

{
    "id": 1,
    "name": "OpenZeppelin",
    "tags": [
	    "tag1",
	    "tag2",
	    "tag3"
    ]
    ...
}
```

- An app can have multiple `tags`
- `tags` can be empty but never `null`
- A `tag` can be part of multiple apps

fmrsabino:
Closes #486 

- Add a new model: `Tag`. Currently it can only contain a `name`
- A `Tag` can be part of multiple `SafeApps` and multiple `SafeApps` can contain the same `Tag`
- A `SafeApp` can have no `Tags` associated with it
- This field is only serialised in the response if `SAFE_APPS_TAGS_FEATURE_ENABLED` is `true`
  * if `SAFE_APPS_TAGS_FEATURE_ENABLED` is enabled then the `tags` field is always returned in the response
  * if `SAFE_APPS_TAGS_FEATURE_ENABLED` is disabled then the `tags` field is not returned

```javascript
GET /api/v1/safe-apps

{
    "id": 1,
    "name": "OpenZeppelin",
    "tags": [
	    "tag1",
	    "tag2",
	    "tag3"
    ]
    ...
}
```


src/safe_apps/migrations/0008_tag.py
```diff
@@ -0,0 +1,32 @@
...
+                (
+                    "id",
+                    models.BigAutoField(
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("name", models.CharField(max_length=255)),
```
katspaugh, line 25: Oh, so it _is_ a model with id and name. The joke's on me :D

src/safe_apps/admin.py
```diff
@@ -23,12 +24,24 @@ def queryset(self, request: Any, queryset: QuerySet[SafeApp]) -> QuerySet[SafeAp
         return queryset
 
 
+class TagInline(admin.TabularInline[Model]):
```
katspaugh, line 27: Will we be able to create tag by typing right in the Safe App input?

https://djangopackages.org/grids/g/tagging/ – looks like there are additional packages for this purpose, or are they not needed?

fmrsabino: > Will we be able to create tags by typing right in the Safe App input?

It'd be similar to what we have for `Feature` and `Wallet` for chains. We can iterate on the admin part as we move along with this feature.

> https://djangopackages.org/grids/g/tagging/ – looks like there are additional packages for this purpose, or are they not needed?

I prefer to not introduce 3rd party dependencies just for this. I would start with what Django offers out of the box.

src/safe_apps/migrations/0008_tag.py
fmrsabino: This is so that a `Tag` can be tied to multiple `SafeApps`. If we need to update a tag for all the apps then we just need to edit in one place.

src/config/settings.py
```diff
@@ -225,3 +225,7 @@
         allowed_csrf_origins.strip()
         for allowed_csrf_origins in allowed_csrf_origins.split(",")
     ]
+
+SAFE_APPS_TAGS_FEATURE_ENABLED = bool(
+    strtobool(os.getenv("SAFE_APPS_TAGS_FEATURE_ENABLED", "false"))
```
Uxio0, line 230: Didn't know about `strtool`, nice @fmrsabino !

src/safe_apps/tests/test_views.py
```diff
@@ -410,6 +440,7 @@ def test_provider_not_returned_in_response(self) -> None:
         self.assertCountEqual(response.json(), json_response)
```
Uxio0, line 440: That's something I should fix also on the other services, but recently I found out that https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertDictEqual is the recommended way to compare dictionaries. Not applicable in this case as you are comparing a list of dictionaries, but something to be aware of 😉 . Maybe you already knew

Uxio0 (approved):

PR #532, opened 2022-05-12
Add url filter for SafeApps

fmrsabino:
See https://github.com/safe-global/safe-client-gateway/issues/855

- Add query parameter to `SafeAppsListView` which matches the `url` field of a `SafeApp`
- The url needs to be an exact match ie.: the protocol, path, any slash, query parameter, etc. need to be the same

```javascript
GET /api/v1/safe-apps/?url=https%3A%2F%2Fapp.aave.com

[
  {
    "id": 1,
    "url": "https://app.aave.com",
    "name": "Aave v2",
    ...
]
```

### DIFF
--- a/src/safe_apps/tests/test_views.py
+++ b/src/safe_apps/tests/test_views.py
@@ -574,3 +574,31 @@ class TagsTests(APITestCase):
         json_response = response.json()
         self.assertEqual(response.status_code, 200)
         self.assertEqual(json_response[0]["tags"], [tag_2.name, tag_1.name])
+
+
+class SafeAppsUrlQueryTests(APITestCase):
+    def test_query_url_match(self) -> None:
+        safe_app = SafeAppFactory.create()
+        url = reverse("v1:safe-apps:list") + f"?url={safe_app.url}"
+
+        response = self.client.get(path=url, data=None, format="json")
+
+        json_response = response.json()
+        self.assertEqual(response.status_code, 200)
+        # There should be a non-empty list
+        self.assertTrue(len(json_response) > 0)
+        # All items should have safe_app.url as the url
+        self.assertTrue(
+            all(map(lambda item: item["url"] == safe_app.url, json_response))
+        )
+
+    def test_query_url_no_match(self) -> None:
+        safe_app = SafeAppFactory.create(url="http://test.com")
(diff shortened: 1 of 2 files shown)


src/safe_apps/views.py
```diff
@@ -29,9 +29,21 @@ class SafeAppsListView(ListAPIView):
         description="Used to filter Safe Apps that are available on `clientUrl`",
         type=openapi.TYPE_STRING,
     )
+    _swagger_url_param = openapi.Parameter(
+        "url",
+        openapi.IN_QUERY,
+        description="Filter Safe Apps available from `url`",
```
mikhailxyz, line 35: nit: would include in the description that the filter looks for an exact match

mikhailxyz (dismissed):

src/safe_apps/views.py
fmrsabino: Yeah that makes sense! Will update it.

fmrsabino: Updated here: [c983aad](https://github.com/safe-global/safe-config-service/pull/532/commits/c983aad10be6553c85cd72a29a4a0a85967fb2bc)

image

mikhailxyz (approved):

src/safe_apps/tests/test_views.py
```diff
@@ -574,3 +574,31 @@ def test_multiple_tags(self) -> None:
...
+
+    def test_query_url_no_match(self) -> None:
+        safe_app = SafeAppFactory.create(url="http://test.com")
+        query_url = safe_app.url + "/"
+        url = reverse("v1:safe-apps:list") + f"?url={query_url}"
+
+        response = self.client.get(path=url, data=None, format="json")
+
+        json_response = response.json()
+        self.assertEqual(response.status_code, 200)
```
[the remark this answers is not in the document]
fmrsabino: > I think that if the queried endpoint doesn't exist or no match the endpoint should returns 404.

In this case the endpoint matches. We have a filter on the `?url` which either returns 0 or multiple results but it should be seen as a successful query.

If the `url` would be on the path and we'd query it like `/api/v1/safe-apps/<url>` then it'd make sense yeah 🙂

PR #554, opened 2022-05-26
Switch production ready to development ready

### LINKED ISSUE #553: Error running docker compose web
**Describe the bug**
When I run ``` docker compose up ``` I get an error with the following message:
```
safe-config-service-web-1    | django.core.exceptions.ImproperlyConfigured: Storage set to S3 but AWS is not configured
```

**To Reproduce**
- Copy the environment ``` cp .env.example .env ```
- Run docker ``` docker compose up ```

**Expected behavior**
Run the service without need a S3 storage. 

**Environment (please complete the following information):**
 - Dev
 - OS: Linux 5.15.32-1-MANJARO
 - Version: last commit in main branch

moisses89:
Closes #553
- Modify .env.sample to development ready. 
- Modify Readme according development ready.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ The environment variables are set via the `.env` file. The configuration in `.en
 cp .env.example .env
 ```
 
-**Important:** It's recommended change properly passwords and another default parameters for production environment.
+**Important:** It is recommended to set sensitive fields (like passwords) and adjust other parameters for your production environment needs. The `.env.sample` configuration shouldn't be used for any production deployment.
 
 ### 2. Running the service image
 


moisses89:

fmrsabino (changes requested):
`docker compose` takes the `.env` file and not the `.env.sample` file.

The `.env` file default is production ready so S3 is used. You can copy the `.env.sample` file to `.env`, edit to your local needs and that'd be it for local development – since the file is not tracked by git you don't need to worry about upstream changes.

Uxio0:
> The `.env` file default is production ready so S3 is used. You can copy the `.env.sample` file to `.env`, edit to your local needs and that'd be it for local development – since the file is not tracked by git you don't need to worry about upstream changes.

@fmrsabino I have to disagree in this part. It's not really common that we assume that `S3` would be the default storage for the service. And even ourselves are not using this `env` file on production. But I think we can agree if we just have something like `.env.sample.prod` and `.env.sample.local`

Uxio0:
> That is up to us to decide. There's no rule on what the `.env.sample` should be (but maybe I'm missing it). For this project: `.env.sample` considers production defaults (as it is explained at the top of the `.env.sample` file) and in the README.

I agree with you on this @fmrsabino . I think the point is that it should be easy to set it up both for new people wanting to contribute to the service that for people wanting to deploy it to production, so we should have at least 2 environment files. I don't think it's fair that someone who wants to contribute to the service has to be messing and understanding all the environment variables (and also doing not very intuitive things like disabling S3 storage) when getting into the project

fmrsabino:
> > The `.env` file default is production ready so S3 is used. You can copy the `.env.sample` file to `.env`, edit to your local needs and that'd be it for local development – since the file is not tracked by git you don't need to worry about upstream changes.
> 
> @fmrsabino I have to disagree in this part. It's not really common that we assume that `S3` would be the default storage for the service. And even ourselves are not using this `env` file on production. But I think we can agree if we just have something like `.env.sample.prod` and `.env.sample.local`

It is common to assume S3 in production for our production needs. i see this on two fronts:

1. We have development defaults – we need to configure more on production
2. We have production defaults – we need to configure more for development

For 1. you really need to make sure that if you use the `.env` file that you are overriding the environment variables required for production. If you miss one then an attack vector could be checking the defaults set for development and targeting a production instance (only true if the `.env` file is used for the configuration).

For 2. You setup a local environment with the production defaults and override each setting as you need for local development purposes. You do this one time while setting up the project.

> I agree with you on this @fmrsabino . I think the point is that it should be easy to set it up both for new people wanting to contribute to the service that for people wanting to deploy it to production, so we should have at least 2 environment files. I don't think it's fair that someone who wants to contribute to the service has to be messing and understanding all the environment variables (and also doing not very intuitive things like disabling S3 storage) when getting into the project

I agree that if this is not clear that it should be made so.

fmrsabino:
We've decided to change the scope of this change:

- `.env.sample` will have the development defaults
- We will still have a single `.env` file
- README should be updated accordingly

Uxio0 (approved):

fmrsabino:
@moisses89 is the PR ready for review? The last commit seemed to be WIP

moisses89:
> @moisses89 is the PR ready for review? The last commit seemed to be WIP

It was needing a review from my side, now is ready and removed WIP in the commit.

rmeissner (dismissed):

README.md
```diff
@@ -13,14 +13,12 @@ The `safe-config-service` is a service that provides configuration information i
 
 ### 1. Configuration
 
-The environment variables are set via the `.env` file. The configuration in `.env.example` is meant to be production ready. You can copy it and adjust it to your development needs (refer to the file for the explanation about each environment variable).
-
+The environment variables are set via the `.env` file. The configuration in `.env.example` is meant to be development ready. You can copy it and adjust it to your development needs (refer to the file for the explanation about each environment variable).
```
fmrsabino, line 16: nitpick: make it more prominent that the `.env.example` should only be used only for development.

```suggestion
The environment variables are set via the `.env` file. The configuration in `.env.example` is meant to be used **only for development**. You can copy it and adjust it to your development needs (refer to the file for the explanation about each environment variable).
```

README.md
```diff
@@ -13,14 +13,12 @@ The `safe-config-service` is a service that provides configuration information i
...
 ```shell
 cp .env.example .env
 ```
 
-**Important:** Some variables are required to be set before running the application: `SECRET_KEY`, `POSTGRES_USER`, `POSTGRES_PASSWORD`.
-They can be set either locally on your environment or (as a provided example) by uncommenting these variables from the `.env` file.
+**Important:** It's recommended change properly passwords and another default parameters for production environment.
```
fmrsabino, line 21: nitpick: some types but also generalising to any field.

```suggestion
**Important:** It is recommended to set sensitive fields (like passwords) and adjust other parameters for your production environment needs. The `.env.sample` configuration shouldn't be used for any production deployment.
```

fmrsabino (commented):
Looks good!

I have 2 more suggestions and we can merge it! 🚀

README.md
moisses89: Yes It is more clear. :)

PR #556, opened 2022-05-26
Add python wheels to dockerfile

### LINKED ISSUE #250: Use python wheels
Wheels is a standard for distributing python packages and should be used in order to avoid running the `setup.py` for some packages among other advantages such as:

> 1. Faster installation for pure Python and native C extension packages.
> 2. Avoids arbitrary code execution for installation. (Avoids setup.py)
> 3. Installation of a C extension does not require a compiler on Linux, Windows or macOS.
> 4. Allows better caching for testing and continuous integration.
> 5. Creates .pyc files as part of installation to ensure they match the Python interpreter used.
> 6. More consistent installs across platforms and machines.

https://pythonwheels.com

moisses89:
Closes #250
- Install python wheels in dockerfile

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,7 +96,7 @@ jobs:
           python-version: '3.10'
       - name: Install dependencies
         run: |
-          pip install -U --no-cache-dir wheel setuptools pip
+          pip install -U --no-cache-dir wheel setuptools
           pip install -r requirements-dev.txt
       - name: Run mypy
         run: mypy --strict src/chains src/safe_apps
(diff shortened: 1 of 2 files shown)


Uxio0 (changes requested):
Can you do that for github actions?

.github/workflows/test.yml
```diff
@@ -96,6 +96,7 @@ jobs:
           python-version: '3.10'
       - name: Install dependencies
         run: |
+          pip install -U --no-cache-dir wheel setuptools pip
```
Uxio0, line 99: Remove `pip` as it's not needed

Uxio0 (changes requested):

Uxio0 (approved):

PR #558, opened 2022-05-26
Enable POSTGRES_PASSWORD in environment example file

moisses89:
Closes #557
- Add POSTGRES_PASSWORD to environment example file

PR #559, opened 2022-05-27
Revert "Uncomment POSTGRES_PASSWORD on environment example file (#558)"

fmrsabino:
This reverts commit 37442835991425874e57027e02d0a07fa2ff0211.

The `.env.sample` file contains production ready defaults. If this needs to change it'd be a bigger change than just this field. See: https://github.com/safe-global/safe-config-service/blob/eca3fa3f5474cbb7a9273592af25fcea8c54ee71/.env.example#L1-L6


Uxio0 (approved):
Agree but we should find another solution

fmrsabino:
> Agree but we should find another solution

@safe-global/safe-services me and @Uxio0 aligned on this. Summary:

- We will have a single `.env.sample` file
- This `.env.sample` file will have the development defaults (instead of the production ones)

PR #685, opened 2022-10-21
Use development settings as default

fmrsabino:
- In order to streamline the development process, the default environment variables were changed from being production-ready to being development-ready
- This includes variables:
  * `SECRET_KEY` – which has the default value of `random-dev-key`
  * `DEFAULT_FILE_STORAGE` – which has the default value of `django.core.files.storage.FileSystemStorage` (AWS setup is no longer required)
  * The Host of the database is now `localhost` – we aim to support mainly running the application locally while the DB can be behind a docker image
  * `DEBUG` is now set to `true` by default
- Rename `.env.example` to `.dev.env` – this file will be used mostly by Docker Compose.
- Overall the goal of the changes that were made were mostly around having the development environment already setup and ready to be worked on

### DIFF
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -25,10 +25,10 @@ BASE_DIR = Path(__file__).resolve().parent.parent
 # See https://docs.djangoproject.com/en/3.2/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = os.getenv("SECRET_KEY", None)
+SECRET_KEY = os.getenv("SECRET_KEY", "random-dev-key")
 
 # SECURITY WARNING: don't run with debug turned on in production!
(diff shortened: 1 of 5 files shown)


src/config/settings.py
```diff
@@ -219,7 +219,7 @@
 # This can be useful if your S3 buckets are public.
 AWS_QUERYSTRING_AUTH = False
 DEFAULT_FILE_STORAGE = os.getenv(
-    "DEFAULT_FILE_STORAGE", "storages.backends.s3boto3.S3Boto3Storage"
+    "DEFAULT_FILE_STORAGE", "django.core.files.storage.FileSystemStorage"
```
hectorgomezv, line 222: I was checking Kubernetes config and I think this is not defined in the service configuration (at least in staging). Could it be using this default value? In that case I suppose we should set this value. Just checking if you're aware of this   🙂

Issue #675, opened 2022-10-13
New "warning" text field in chain configs


katspaugh:
**Is your feature request related to a problem? Please describe.**
We sometimes show a green banner on top of the website when some service is down. This is currently done via Beamer, which is a 3rd party service that we load only if the user agreed to store "support & updates" cookies, which is not the default.

**Describe the solution you'd like**
It would be nice to show this banner natively instead, via the website itself.

I suggest to add a new text field per each chain config, e.g. `warning`, that the app would read and render as a banner for that chain.

<img width="1468" alt="Screenshot 2022-10-13 at 13 55 15" src="https://user-images.githubusercontent.com/381895/195589609-6ddea46a-3634-4910-8dd8-f5ef93a5d265.png">

_This is not urgent, just a potential improvement when we have spare time_

hectorgomezv:
Open for discussion [here](https://www.notion.so/gnosis-safe/Config-Service-UI-Banner-feature-discussion-abfe48ccd366444298421313191524ea)

katspaugh:
Reopened to track it, it's still important to have a way to communicate annoucnement to users w/o a cookie-gated 3rd-party service.

PR #688, opened 2022-10-25
Implement warnings

hectorgomezv:
Relates to #675 

This PR aims to create a new `warning` field for each `Chain`. This new `warning` field is nullable and should be updatable by superusers, and also for support collaborators. That's why a new `support` group is created. The idea is to make all `Chain` fields read-only for support collaborators, except `warning`, so this group can only edit the `warning` field.

Using this approach we avoid the usage of third party libraries or external dependencies. It's implemented by just using Django features. Feedback is welcomed! 🙂


src/chains/models.py
```diff
@@ -119,6 +119,7 @@ class RpcAuthentication(models.TextChoices):
     recommended_master_copy_version = models.CharField(
         max_length=255, validators=[sem_ver_validator]
     )
+    warning = models.CharField(max_length=255, blank=True, null=True)
```
Uxio0, line 122: Maybe a `TextField`?

src/chains/migrations/0038_auto_20221024_1543.py
```diff
@@ -0,0 +1,22 @@
...
+
+from django.apps.registry import Apps
+from django.contrib.auth.models import Group, Permission
+from django.db import migrations
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+
+def create_support_group(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    support_group, created = Group.objects.get_or_create(name="support")
+    permission = Permission.objects.filter(codename="change_chain")
```
Uxio0, line 11: Where is this permission defined? It should be on `Meta` for the `Chain` model: https://docs.djangoproject.com/en/4.1/topics/auth/customizing/#custom-permissions

src/chains/migrations/0038_auto_20221024_1543.py
```diff
@@ -0,0 +1,22 @@
...
+    if permission:
+        support_group.permissions.add(permission[0])
```
Uxio0, line 13: If you only expect to get one permission you can do `Permission.objects.get(codename="change_chain")`

src/chains/migrations/0038_auto_20221024_1543.py
```diff
@@ -0,0 +1,22 @@
...
+from django.apps.registry import Apps
+from django.contrib.auth.models import Group, Permission
+from django.db import migrations
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+
+def create_support_group(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    support_group, created = Group.objects.get_or_create(name="support")
+    permission = Permission.objects.filter(codename="change_chain")
+    if permission:
```
Uxio0, line 12: Permission must exist we shouldn't check this

Uxio0 (changes requested):

src/chains/admin.py
```diff
@@ -38,6 +41,15 @@ class ChainAdmin(admin.ModelAdmin[Chain]):
     )
     inlines = [FeatureInline, GasPriceInline, WalletInline]
 
+    def get_readonly_fields(
+        self, request: HttpRequest, obj: Optional[Chain] = None
+    ) -> List[str]:
+        if request.user.groups.filter(name="support"):  # type: ignore
```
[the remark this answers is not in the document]
fmrsabino: What about having a new table for the warning messages? ChainId + Message.

The endpoint can return the message from this table. The permission then can be done on the table level. Maybe it's easier and simpler this way wdyt?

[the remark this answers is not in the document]
Uxio0: I would prefer not if it's possible to do it in another way, I would prefer to have a normalized database if possible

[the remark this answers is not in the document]
fmrsabino: I was thinking more about this field and in a way it seems that it's not really part of the chain metadata so that's why I thought about a separate table. Even though it's possible in a different way the Chain model itself already has quite some fields. But just an idea (nothing that would block the feature).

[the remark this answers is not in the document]
hectorgomezv: I was thinking in having a separated table as well, it could be nice to keep an history of the changes as well, but I initially discarded it because it seems a bit overkill for the feature. 

About the custom permission... it was my initial approach as well, but in this case we need to still grant edit permission over all Chain fields to superusers, and only over `warning` field for a selected group of users. I thinking in creating an additional permission, something like `can_change_all_chain_fields` and when a user doesn't have it, show just the `warning` as editable. Do you see a simpler way here?

src/chains/models.py
hectorgomezv: Yes, thanks! This is something I forgot to annotate in the PR, but this made me doubt as well 👍

src/chains/migrations/0038_auto_20221024_1543.py
hectorgomezv: I was getting an error when running tests due to this, that's why I included the condition. Maybe it's related with not having the permission defined in `Meta` as you said in [the previous comment](https://github.com/safe-global/safe-config-service/pull/688#discussion_r1004205828)...

Uxio0: That's it. Permission should be created in a migration when adding it to `Meta`, and you can use that migration to add it to the group. Also, if possible, do everything in the same migration

src/chains/admin.py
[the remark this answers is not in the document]
Uxio0: I think I'm missing something, but why not using `if request.user.has_perm("app.permission_name")`?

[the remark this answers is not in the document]
hectorgomezv: Maybe it's me the one missing something, but we need to avoid _support_ group from modifying any field except for `warning` right? So it's kind of _negative permission_. I mean, all the other fields are the ones to restrict to some users, right?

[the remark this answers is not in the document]
Uxio0: I see it more like:
- By default users don't have any permission unless you make them superusers or give them permissions to every table individually.
- `support` group will grant the `edit chains`  and `change_only_warning` to all his members.
- To edit the chains table users will need the `edit chains` permission
- People with `change_only_warning` permission will be able to modify only `warning` field, so `if request.user.has_perm(`change_only_warning`):` every field but warning will be read_only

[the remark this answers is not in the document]
hectorgomezv: @Uxio0 thanks for your feedback and guidance on this 🙂. I was doing some tests and it seems that when creating a new permission (`change_only_warning` in this case), that permission is automatically inherited by superuser(s). So, in this case, that converts all fields in `read_only` for the superusers also (except `warning` of course) 🤔 I see rwo options here:
- Rely on the group
- Have an additional check to verify the user is not a `superadmin` before marking the fields as read-only.

```python
if (
        request.user.has_perm("change_only_warning")
        and not request.user.is_superuser
    ):
```

I think the second option is better, what do you think?

src/chains/migrations/0038_auto_20221024_1543.py
hectorgomezv: @Uxio0 I was testing a lot trying to understand why the migration was failing (because the permission wasn't found), I think the reason why we need `Permission.objects.get_or_create` here instead of simply `Permission.objects.get` is because the permissions data is not inferred/written to the database until all migrations are executed (using the [post_migrate](https://docs.djangoproject.com/en/4.1/ref/signals/#post-migrate) signal if I'm not wrong.

src/chains/admin.py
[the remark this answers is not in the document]
hectorgomezv: I implemented the second option, of course if you see any concerns with it let me know 🙂

src/chains/migrations/0038_auto_20221024_1543.py
hectorgomezv: I think [this issue](https://code.djangoproject.com/ticket/23422) is related. This also explains why test were failing in my local setup (I understand the whole list of migrations gets executed against a test database when running the tests).

src/chains/admin.py
```diff
@@ -38,6 +41,18 @@ class ChainAdmin(admin.ModelAdmin[Chain]):
     )
     inlines = [FeatureInline, GasPriceInline, WalletInline]
 
+    def get_readonly_fields(
+        self, request: HttpRequest, obj: Optional[Chain] = None
+    ) -> List[str]:
+        if (
+            request.user.has_perm("chains.change_only_warning")
+            and not request.user.is_superuser
```
fmrsabino, line 49: Why don't we allow a superuser to change this field? (even without the permission)

src/chains/tests/test_views.py
```diff
@@ -195,6 +196,7 @@ def test_json_payload_format(self) -> None:
             "recommendedMasterCopyVersion": chain.recommended_master_copy_version,
             "disabledWallets": [],
             "features": [],
+            "warning": None,
```
fmrsabino, line 199: We are missing a test if the warning is set.

The factory method should also be updated

fmrsabino (changes requested):

src/chains/migrations/0037_chain_warning.py
```diff
@@ -0,0 +1,46 @@
...
+from django.contrib.auth.models import Group, Permission
+from django.contrib.contenttypes.models import ContentType
+from django.db import migrations, models
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+from ..models import Chain
+
+
+def create_support_group(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    support_group, created = Group.objects.get_or_create(name="support")
```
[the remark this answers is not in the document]
Uxio0: I think having the group created and added trhough a migration makes it more clear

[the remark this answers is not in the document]
fmrsabino: Alright. No strong opinion on this one 👍

src/chains/models.py
```diff
@@ -129,6 +130,9 @@ def get_disabled_wallets(self) -> QuerySet["Wallet"]:
     def __str__(self) -> str:
         return f"{self.name} | chain_id={self.id}"
 
+    class Meta:
+        permissions = [("change_only_warning", "Can change only warning")]
```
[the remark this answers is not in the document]
hectorgomezv: `only` is included in the name because of the nature of this permission. This permission will imply two things for an user: 
- The user can edit the `warning` field.
- The user cannot edit any other `Chain` field, i.e. the rest of the field will be marked as read-only.

I know it is not very intuitive, but we agreed on taking this approach to avoid using a third party dependency which would let us define field-level permissions. Even using that dependency, I suspect the outcome wouldn't be very intuitive either, because our goal is not to grant new permissions, but preventing some users to access some fields.

src/chains/admin.py
hectorgomezv: (Please refer to the [previous comment](https://github.com/safe-global/safe-config-service/pull/688#discussion_r1018095602) for reference)

This function (`get_readonly_fields`) returns the read-only fields, so for an user which have the `change_only_warning` and is not superadmin, will return all the `Chain` fields except `warning`. For superusers, it will return `[]`, so all the fields are editable. In short, we allow superusers to change this field.

(I know this code is not super intuitive to read, please let me know if you find a better approach).

src/chains/tests/test_views.py
hectorgomezv: Added in [02e17b2](https://github.com/safe-global/safe-config-service/pull/688/commits/02e17b2aadef9298675f6238cfaf769509feb33c) 👍🏻

src/chains/models.py
hectorgomezv: Changed in [02e17b2](https://github.com/safe-global/safe-config-service/pull/688/commits/02e17b2aadef9298675f6238cfaf769509feb33c)

hectorgomezv:
I'm drafting this PR since there is an ongoing discussion [in Notion](https://www.notion.so/gnosis-safe/Config-Service-UI-Banner-feature-discussion-abfe48ccd366444298421313191524ea).

hectorgomezv:
I'm closing this PR because [the source issue](https://github.com/safe-global/safe-config-service/issues/675) is also closed now. Please reach out if you want to discuss this feature further 🙂

PR #690, opened 2022-10-31
Set `latest` docker tag on releases

Uxio0:
- We took a look on the different official dockerhub images (https://hub.docker.com/search?q=&type=image&image_filter=official) and they all follow the same "de facto" standard of setting `latest` to the last release of the application.
- They use `stable` for older versions or LTS, and in our case we don't have that and we want users running the last versions of our services.
- With this `latest` tag it will be easy for users to be updated and not having to manually update versions (like in https://github.com/safe-global/safe-infrastructure)
- Also make docker staging/release pushes different jobs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -114,10 +114,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: coveralls --service=github
 
-  docker-publish:
-    if: |
-      (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
-      (github.event_name == 'release' && github.event.action == 'released')
+  docker-publish-staging:
+    if: (github.event_name == 'push' && github.ref == 'refs/heads/main')
     needs: [ flake8, isort, black, django-check ]
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


fmrsabino:
@Uxio0 I'd add a description why this is needed

.github/workflows/test.yml
```diff
@@ -154,6 +154,16 @@ jobs:
...
+      - name: Build and push latest tag
+        id: docker_build_latest
+        if: github.event_name == 'release' && github.event.action == 'released'
+        uses: docker/build-push-action@v3.2.0
+        with:
+          context: .
+          push: true
+          tags: gnosispm/safe-config-service:latest
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new
```
fmrsabino, line 166: Maybe it makes more sense to split the jobs (release vs staging).

Before we could just set the `DOCKER_TAG` but now the behaviour is actually different:
- For releases we push two tags (`latest` and the release)
- For commits we push one tag (`staging`)

.github/workflows/test.yml
```diff
@@ -154,6 +154,16 @@ jobs:
           tags: gnosispm/safe-config-service:${{ env.DOCKER_TAG }}
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache-new
+      - name: Build and push latest tag
+        id: docker_build_latest
```
fmrsabino, line 158: The `id` seems to be used to print the Image digest at the end. If this is an optional field maybe we can just remove it from here

.github/workflows/test.yml
Uxio0: Done @fmrsabino !

fmrsabino (approved):
Awesome! I think is' more clear when looking at the pipeline! 👏

PR #698, opened 2022-11-04
Reject null input for safe apps filters

### LINKED ISSUE #697: 500 Error on unsanitized URL query params
**Describe the bug**
Error response with 500 Internal server Error is returned to the clients when a unsanitized URL query param is sent to the service.

**To Reproduce**
- Check: https://safe-config.safe.global/api/v1/safe-apps/?url=%00

**Expected behavior**
URL input is sanitized beforehand.

**Environment**
 - Staging & production
 - All chains

Uxio0:
- Closes #697

I provided a quick PR for this, but the right solution would be using Django filtering (I don't know if it's worth it to migrate now): https://www.django-rest-framework.org/api-guide/filtering/

As you can see this protection is provided by default: https://github.com/encode/django-rest-framework/blob/c10f2266222c434485889b08cc1463acdb8fa169/rest_framework/fields.py#L739

### DIFF
--- a/src/safe_apps/tests/test_views.py
+++ b/src/safe_apps/tests/test_views.py
@@ -267,6 +267,16 @@ class FilterSafeAppListViewTests(APITestCase):
         self.assertEqual(response.status_code, 200)
         self.assertCountEqual(response.json(), json_response)
 
+    def test_apps_returned_on_null_client_url(self) -> None:
+        url = reverse("v1:safe-apps:list") + "?clientUrl=\0"
+        response = self.client.get(path=url, data=None, format="json")
+        self.assertEqual(response.status_code, 200)
+
+    def test_apps_returned_on_null_url(self) -> None:
+        url = reverse("v1:safe-apps:list") + "?url=\0"
+        response = self.client.get(path=url, data=None, format="json")
+        self.assertEqual(response.status_code, 200)
+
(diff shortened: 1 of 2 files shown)


src/safe_apps/tests/test_views.py
```diff
@@ -267,6 +267,17 @@ def test_apps_returned_on_non_existent_client_url(self) -> None:
...
+        url = reverse("v1:safe-apps:list") + "?clientUrl=\0"
+
+        response = self.client.get(path=url, data=None, format="json")
+        self.assertEqual(response.status_code, 200)
+
+        url = reverse("v1:safe-apps:list") + "?url=\0"
+
+        response = self.client.get(path=url, data=None, format="json")
+        self.assertEqual(response.status_code, 200)
+
```
fmrsabino, line 280: Nitpick: I would just split this in two tests (one for `clientUrl` and another one for `url`) – so that we get independent feedback on the status of both checks.

fmrsabino (approved):

hectorgomezv (approved):

src/safe_apps/tests/test_views.py
Uxio0: Good suggestion, done

fmrsabino (approved):

Issue #517, opened 2022-04-27
Refactor the project structure to follow the same from other Safe projects


Uxio0:
**Is your feature request related to a problem? Please describe.**
Structure of the project does not follow any standard.

**Provide references to the feature you are implementing that requires this change**
Project should follow the other Safe backend projects structure (like `tx-service`)

**Describe the solution you'd like**
- Use https://github.com/cookiecutter/cookiecutter-django#usage structure
Most relevant changes:
- Rename `src` folder
- Create folder for configuration
- Move `conftest.py` and `manage.py` to root folder
- Use full naming for Django apps (prepending the project name)

**Describe alternatives you've considered**

**Additional context**

PR #699, opened 2022-11-04
Configure and run pre-commit

Uxio0:
- Related to #517


.github/workflows/test.yml
```diff
@@ -12,51 +12,22 @@ on:
     types: [ released ]
 
 jobs:
-  flake8:
-    name: flake8
+  linting:
```
fmrsabino, line 15: What's the reason to change this?

.github/workflows/test.yml
```diff
@@ -12,51 +12,22 @@ on:
...
-          python-version: '3.10'
-      - run: python -m pip install isort
-      - name: isort
-        uses: liskin/gh-problem-matcher-wrap@v1
-        with:
-          linters: isort
-          run: isort --check --profile black src/
+    strategy:
+      matrix:
+        python-version: ["3.11"]
```
fmrsabino, line 19: Why change it to python 3.11? We haven't tested its support yet

.github/workflows/test.yml
```diff
@@ -12,51 +12,22 @@ on:
...
         with:
-          python-version: '3.10'
-      - run: python -m pip install black
-      - name: black
-        run: black --check src/
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: pip install pre-commit
+      - name: Run pre-commit
+        run: pre-commit run --all-files
```
fmrsabino, line 30: Isn't this changing the source files?

Pre-commit should also be used optionally, locally for committing purposes in my opinion. The CI then validates the checks independently.

.pre-commit-config.yaml
```diff
@@ -1,15 +1,28 @@
...
+      - id: check-docstring-first
+      - id: check-merge-conflict
+      - id: debug-statements
+      - id: detect-private-key
+      - id: requirements-txt-fixer
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+        types: [python]
+      - id: check-yaml
+      - id: check-added-large-files
```
fmrsabino, line 28: Why are we adding all of these?

requirements-dev.txt
```diff
@@ -2,16 +2,16 @@
...
 flake8==5.0.4
 isort==5.10.1
-pre-commit==2.20.0
-pytest-django==4.5.2
-responses==0.22.0
 
 # mypy and PEP 561 stubs
 mypy==0.982
-django-stubs==1.12.0
-djangorestframework-stubs==1.7.0
```
fmrsabino, line 16: I split the python stubs for readability. Why are we changing this?

fmrsabino (changes requested):
This is doing many changes and I believe more context is needed

.github/workflows/test.yml
Uxio0: Sorry this was not intended

.github/workflows/test.yml
Uxio0: This is indeed modifying source files is something is wrong, but will abort the execution.

I think `precommit` has multiple checks that were not run separately (for example removing trailing spaces). I think `precommit` should not be optional and mandatory for everyone collaborating to this project so it's more convenient to just run in on the CI instead of configuring multiple checks separately

.pre-commit-config.yaml
Uxio0: We are already using them for the transaction service and we should use it in our python services. Those are very useful checks

requirements-dev.txt
Uxio0: precommit is sorting it lexicographically

Uxio0:
@hectorgomezv do you think this is relevant nowadays or should I close it? I still feel that all the Safe projects should have the same linting/style

hectorgomezv:
> @hectorgomezv do you think this is relevant nowadays or should I close it? I still feel that all the Safe projects should have the same linting/style

Sorry @Uxio0, I completely missed this comment. I've fixed some conflicts on the PR. If you are OK with the changes I think we can merge.

hectorgomezv (approved):

hectorgomezv:
@Uxio0 we did some amendments (branch rebasing mostly) to the PR, if you are OK with the current status we will merge it finally 🙂

.github/workflows/test.yml
```diff
@@ -12,78 +12,22 @@ on:
...
-          pip install -U wheel setuptools
-          pip install -r requirements-dev.txt
-      - name: flake8
-        uses: liskin/gh-problem-matcher-wrap@v3
-        with:
-          linters: flake8
-          run: flake8
+    strategy:
+      matrix:
+        python-version: ["3.12.2"]
```
Uxio0, line 19: I would leave this as `3.12` so hotfixes are auto updated

Uxio0 (commented):
Just a minor comment, but everything else looks good!

hectorgomezv:
> Just a minor comment, but everything else looks good!

Thanks! Done in [2b6a52b](https://github.com/safe-global/safe-config-service/pull/699/commits/2b6a52bf3899accbdb4372c45bbf9907acf839cd)

PR #731, opened 2022-12-09
docs: fix docker permission denied error

gzliudan:
The below error would occur when run `docker compose up -d db` as a nonprivileged user:
```
permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/json?all=1&filters=%7B%22label%22%3A%7B%22com.docker.compose.project%3Dsafe-config-service%22%3Atrue%7D%7D": dial unix /var/run/docker.sock: connect: permission denied
```

We should use sudo to run this command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ pip install -r requirements-dev.txt
 #### 2. Launch the Postgres database image
 
 ```shell
-docker compose up -d db
+sudo docker compose up -d db
 ```
 
 #### 3. Execute pending database migrations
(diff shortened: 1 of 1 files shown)


fmrsabino:
This is up to the configuration that you have in your machine and respective permissions for the user in place.

The `README` is not intended to be a global solution so we cannot assume that the user also has `sudo` installed.

gzliudan:
After put the user into docker group, no need to use sudo.

Issue #755, opened 2023-01-12
Store Safe App features


mikhailxyz:
**Is your feature request related to a problem? Please describe.**
As a part of improving the [discoverability of Safe Apps](https://www.notion.so/gnosis-safe/Discoverability-of-Safe-Apps-ae7f9a4066d64a45ac038397f1703511), we want to highlight applications that have safe specific features such as batched transactions

**Provide references to the feature you are implementing that requires this change**
- [Problem statement](https://www.notion.so/gnosis-safe/Discoverability-of-Safe-Apps-ae7f9a4066d64a45ac038397f1703511)
- [Figma](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=2231%3A64547&t=IgOfrzNdC4yo5nFc-3)

**Describe the solution you'd like**
The safe apps model would have an array of features similar to features on the chain model

**Describe alternatives you've considered**
At the moment, the app can only enable a single safe-specific feature (batched transactions) and we considered adding a simple boolean field but an array of features sounds like a more long-term oriented solution

**Additional context**
I'm willing to work on this problem myself :)

fmrsabino:
@mikhailxyz This can be closed right (as the feature was merged)?

mikhailxyz:
yes, thanks for the reminder @fmrsabino

PR #756, opened 2023-01-13
Feature: add features to safe apps

mikhailxyz:
This PR:
- Implements https://github.com/safe-global/safe-config-service/issues/755 by creating a new table for safe app features with a many to many relationship


mikhailxyz:

src/safe_apps/migrations/0009_safe_apps_features.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 4.1.5 on 2023-01-12 16:35
+from django.apps.registry import Apps
+from django.db import migrations, models
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+from django.db.migrations import RunPython
+
+
+def add_default_features(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    Feature = apps.get_model("safe_apps", "Feature")
+    Feature.objects.create(key="BATCHED_TRANSACTIONS")
```
fmrsabino, line 10: I'm not sure if the migration itself warrants a single default Safe App `Feature`. Given that this project is also open-source I'd probably leave our own use-case out from the migration and add it via the admin. Wdyt?

src/safe_apps/serializers.py
```diff
@@ -35,11 +35,18 @@ def to_representation(self, instance: Tag) -> str:
...
+    def to_representation(instance: Feature) -> str:
+        return instance.key
+
+
 class SafeAppsResponseSerializer(serializers.ModelSerializer[SafeApp]):
     id = serializers.IntegerField(source="app_id")
     provider = ProviderSerializer()
     access_control = serializers.SerializerMethodField()
     tags = serializers.SerializerMethodField()
+    features = serializers.SerializerMethodField()
```
fmrsabino, line 49: This will expose `features` once this hits staging and then production. If there are no concerns in its release from your side then I think it's good to go. If you consider this as something that is work in progress we can hide this behaviour with a feature flag (so that you can test on staging freely without affecting the API exposed by production).

fmrsabino (dismissed):
👏 🚀

src/safe_apps/serializers.py
mikhailxyz: the data format has been agreed and we do not expect any changes in the near future, so i think it's good to go without hiding

fmrsabino (approved):
🚀 🚀 🚀

PR #764, opened 2023-01-17
Add meta `ref_name` to `Feature` serializers

mikhailxyz:
Fixes swagger not loading because of Serializer name conflict

### DIFF
--- a/src/safe_apps/serializers.py
+++ b/src/safe_apps/serializers.py
@@ -36,6 +36,9 @@ class TagSerializer(serializers.Serializer[Tag]):
 
 
 class FeatureSerializer(serializers.Serializer[Feature]):
+    class Meta:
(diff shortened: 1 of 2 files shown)


src/safe_apps/serializers.py
```diff
@@ -36,6 +36,9 @@ def to_representation(self, instance: Tag) -> str:
 
 
 class FeatureSerializer(serializers.Serializer[Feature]):
+    class Meta:
+        ref_name = "SafeApp.Feature"
```
fmrsabino, line 40: Similar to the comment above:

```suggestion
        ref_name = "safe_apps.serializers.FeatureSerializer"
```

fmrsabino (changes requested):
Two small suggestions

fmrsabino (approved):
🚀 🚀 🚀

Issue #739, opened 2022-12-13
Store social media links for Safe Apps


mikhailxyz:
**Is your feature request related to a problem? Please describe.**
As a part of improving the [discoverability of Safe Apps](https://www.notion.so/gnosis-safe/Discoverability-of-Safe-Apps-ae7f9a4066d64a45ac038397f1703511), we want to have links to app's websites and social media platofrms.

**Provide references to the feature you are implementing that requires this change**
- [Problem statement](https://www.notion.so/gnosis-safe/Discoverability-of-Safe-Apps-ae7f9a4066d64a45ac038397f1703511)
- [Figma](https://www.figma.com/file/ptTs6lDBeUuLNySroJ5PiF/Web-Master-File?node-id=2231%3A64547&t=IgOfrzNdC4yo5nFc-3)

**Describe the solution you'd like**
The safe apps model would have an array of social links with an URL and link type

**Describe alternatives you've considered**
Fetching them from coingecko/coinmarketcap/etc

**Additional context**
I'm willing to work on this problem myself :)

johannesmoormann:
Please wait for prioritization meeting results before committing to the implementation.

fmrsabino:
> The safe apps model would have an array of social links with an URL and link type

Do we want to target specific social links or do we want to have a generic collection of social media links? Do they have an icon? (I can't open the Figma link)

> The safe apps model would have an array of features similar to features on chain model

Can you give an example of what these features represent?

---

I would consider splitting this ticket into two. I don't think they are related to each other right? One for media links and the other for features.

mikhailxyz:
> Do we want to target specific social links or do we want to have a generic collection of social media links? Do they have an icon? (I can't open the Figma link)

<img width="737" alt="Screenshot 2023-01-12 at 12 19 38" src="https://user-images.githubusercontent.com/16622558/212053982-b6ea8c61-acde-40af-bab4-edd8696e45d0.png">

This is how it looks. We want to link application's social media platform such as twitter, github, discord, landing page, etc

mikhailxyz:
> Can you give an example of what these features represent?

Right now the only feature the app can enable is batched transactions. Perhaps it can be a simple boolean field on the apps model, but i think an array with features is a more long-term solution 

> I would consider splitting this ticket into two. I don't think they are related to each other right? One for media links and the other for features.

Correct, they're unrelated but since they both seemed small and straightforward I included both here. I can create a separate ticket for features if you say it's better

[...]

PR #765, opened 2023-01-17
Feature: Add social profile links to Safe Apps

mikhailxyz:
This PR:
- Implements #739


mikhailxyz:
I need to also include links to the admin page, otherwise the PR is good to review

src/safe_apps/tests/test_views.py
```diff
@@ -690,3 +771,39 @@ def test_query_url_no_match(self) -> None:
...
+        response = self.client.get(path=url, data=None, format="json")
+
+        json_response = response.json()
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(json_response[0]["socialProfiles"], [])
+
+    def test_multiple_social_profiles(self) -> None:
+        safe_app = SafeAppFactory.create()
+        social_profile_1 = SocialProfileFactory.create(platform="Z", safe_app=safe_app)
+        social_profile_2 = SocialProfileFactory.create(platform="A", safe_app=safe_app)
```
fmrsabino, line 790: Since the factories don't perform a `full_clean` the `SocialProfile` will be successfully created even though the platforms are not part of our `TextChoices`.

We should test that by using the respective social networks that we support that these are currently rendered on the client side. That includes things like:
- If the text choice is `DISCORD` that `DISCORD` is correctly serialized as a string in the payload (and the same of other networks)
- How do we render Safe Apps with multiple social profiles of the same type.
- How do we render Safe Apps with multiple social profiles of different types.


However, `TextChoices` are not validated by default so that's why this test works right now. I suggest adding something to the Factory like:

```python
@factory.post_generation
def validate(obj, create, extracted, **kwargs):
    if not create:
        return
    obj.full_clean()
```

By doing this, this test should fail.

src/safe_apps/tests/test_models.py
```diff
@@ -70,3 +71,12 @@ class FeatureTestCase(TestCase):
     def test_str_method_outputs_feature_key(self) -> None:
         feature = FeatureFactory.create()
         self.assertEqual(str(feature), f"Safe App Feature: {feature.key}")
+
+
+class SocialProfileTestCase(TestCase):
+    def test_str_method_outputs_social_profile_name(self) -> None:
```
fmrsabino, line 77: We should also test what happens when you save a model with invalid `TextChoices`.

And also that a Social Profile cannot exist without a `SafeApp`.

fmrsabino (changes requested):

src/safe_apps/tests/test_views.py
mikhailxyz: Thanks! Good catches, have just addressed them in my new commit

fmrsabino (approved):
Great work 👏 🚀

PR #766, opened 2023-01-18
Feature: Add Safe App features to admin

mikhailxyz:
I forgot to include features to admin in https://github.com/safe-global/safe-config-service/pull/756, this PR adds it

### DIFF
--- a/src/safe_apps/admin.py
+++ b/src/safe_apps/admin.py
@@ -3,7 +3,7 @@ from typing import Any
 from django.contrib import admin
 from django.db.models import Model, QuerySet
 
-from .models import Client, Provider, SafeApp, Tag
+from .models import Client, Feature, Provider, SafeApp, Tag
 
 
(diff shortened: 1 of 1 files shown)


fmrsabino (dismissed):
LGTM. `mypy` reported some issues though

mikhailxyz:
@fmrsabino ahh thought it was a part of the pre-commit hook

fmrsabino:
> @fmrsabino ahh thought it was a part of the pre-commit hook

It's a little bit more tricky to run `mypy` with a `pre-commit` hook because `mypy` relies on a virtual environment with some dependencies that we explicitly add: https://github.com/safe-global/safe-config-service/blob/b530ee022914367ffde4403e18bce60e8b1ddb8c/requirements-dev.txt#L13-L17

Since there's some setup involved with `mypy` we decided not to integrate this with pre-commit hooks as the value that it would provide would also be minimal.

fmrsabino (approved):

PR #835, opened 2023-04-19
Add hidden field to Chain model

### LINKED ISSUE #833: Add a property for Chain to hide
**Background**
- Adding a chain is a very tedious process
- We need to add chains temporary to do proper QA before releasing
- After adding a chain, removing and doing QA it could be possible that when we add it again we make a mistake and chain is not working as expected

- A `hidden` field that will remove a `Chain` from every result

hectorgomezv:
Closes #833 

This PR adds the `hidden` property to `Chain` model, with its default value set to `False`. This property makes it possible to mark a chain as hidden, so it can be created/configured beforehand. A chain will be only visible through the API when this `hidden` property is set to `False`.

### DIFF
--- a/src/chains/migrations/tests/test_migration_0037.py
+++ b/src/chains/migrations/tests/test_migration_0037.py
@@ -1,0 +1,67 @@
+from django.db.migrations.state import StateApps
+
+from chains.migrations.tests.utils import TestMigrations
+
+
+class Migration0034TestCase(TestMigrations):
+    migrate_from = "0036_alter_chain_transaction_service_uri_and_more"
+    migrate_to = "0037_chain_hidden"
+
+    def setUpBeforeMigration(self, apps: StateApps) -> None:
+        Chain = apps.get_model("chains", "Chain")
+        Chain.objects.create(
+            id=1,
+            name="Mainnet",
+            short_name="eth",
+            description="",
+            l2=False,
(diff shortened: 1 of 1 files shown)


Uxio0 (changes requested):
I would add one test and update the factories with the hidden field: https://github.com/safe-global/safe-config-service/blob/main/src/chains/tests/factories.py#L14

fmrsabino:
> I would add one test and update the factories with the hidden field: https://github.com/safe-global/safe-config-service/blob/main/src/chains/tests/factories.py#L14

Also maybe a migration test to make sure that previous available chains are visible once the migrations is applied.

hectorgomezv:
> I would add one test and update the factories with the hidden field: https://github.com/safe-global/safe-config-service/blob/main/src/chains/tests/factories.py#L14

I've added the `hidden` property to the `ChainFactory` class, and also some tests verifying the hidden chains get ignored by the list response and the client gets a 404 when requesting the detail view of a hidden chain in [989ea03](https://github.com/safe-global/safe-config-service/pull/835/commits/989ea0308e8d33cffadf531879fde191a85fd77f)

Uxio0 (approved):

Uxio0:
Looks good to me, thanks for implementing this @hectorgomezv !

PR #856, opened 2023-05-24
Support base url with prefix

suraneti:
If I set `CGW_URL=http://localhost:8000/cgw`, `/cgw` will be removed and cause a 404 error.

http://localhost:8000/v2/flush (404 NOT FOUND)

### DIFF
--- a/src/clients/safe_client_gateway.py
+++ b/src/clients/safe_client_gateway.py
@@ -16,6 +16,12 @@ def setup_session() -> requests.Session:
     session.mount("https://", adapter)
     return session
 
+def valid_url(url) -> bool:
+    try:
+        result = urlparse(url)
+        return all([result.scheme, result.netloc])
+    except ValueError:
+        return False
 
 def flush(
     cgw_url: Optional[str], cgw_flush_token: Optional[str], json: Dict[str, Any]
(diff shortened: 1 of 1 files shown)


src/clients/safe_client_gateway.py
```diff
@@ -16,18 +16,27 @@ def setup_session() -> requests.Session:
     session.mount("https://", adapter)
     return session
 
+def valid_url(url) -> bool:
+    try:
+        result = urlparse(url)
+        return all([result.scheme, result.netloc])
+    except ValueError:
+        return False
```
fmrsabino, line 24: I think this can be misleading as we `urlparse` does not do url validation.

Maybe we can be more specific on what we are trying to achieve here which is to update correctly the path.

So something like `def append_flush_path(url) -> str` which would extend the `url` path to have `/v2/flush`.

What do you think of the following?

```python
def append_flush_path(url) -> str:
    parts = list(urlsplit(url))
    # parts[2] corresponds to the url path
    # https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit
    parts[2] += "/v2/flush"
    return urlunsplit(parts)
```

and then we'd use like this:

```python
try:
    url = append_flush_path(cgw_url)
except ValueError:
    logger.error("CGW_URL is not a valid url")
    return
```

fmrsabino: On a second thought maybe just appending the path to the string-url would be enough. The main reason is that neither `urlparse` or `urlsplit` validate the `url`. Given that we can build the url with the following:

```python
url = cgw_url.rstrip('/') + "/v2/flush"
```

hectorgomezv:
I'm closing this due to inactivity + the changes in the PR affect code that was removed from the project (the service does not use a `flush` endpoint anymore).

PR #900, opened 2023-07-06
Support a second Client Gateway URL

### LINKED ISSUE #898: Support a second Client Gateway webhook URL
***Context***

A second target URL for webhook triggered by service events (chain configuration change, new chain, deleted chain, etc.) can be used during canary deployments, so both the production Client Gateway and the canary one receive webhooks and they can clear the cache accordingly.

***Describe the solution you'd like (short-term)***

To define the Client Gateway webhooks target the service only takes a `CGW_URL` environment variable. At least, a second target is needed to use a canary deployment strategy.  So:
- `ALTERNATIVE_CGW_URL` environment variable should be defined in the service.
- The service should send webhooks to both `CGW_URL` and `ALTERNATIVE_CGW_URL` urls.

***Describe the solution you'd like (long-term, out of reach ATM)***

Several webhook targets can be defined using the Django Admin. This is not covered by this issue but would be nice to have it in the future.

hectorgomezv:
Closes #898 

This PR:
- Adds `ALTERNATIVE_CGW_URL` and `ALTERNATIVE_CGW_FLUSH_TOKEN` as environment variables to the service settings.
- If these variables are set, it makes the service execute a second HTTP request to `${ALTERNATIVE_CGW_URL}/v2/flush`, in order to invalidate the cache in a second Client Gateway pointed by that URL.
- Modifies the `test_signals` unit tests accordingly to verify the expected requests are executed.

### DIFF
--- a/src/safe_apps/tests/test_signals.py
+++ b/src/safe_apps/tests/test_signals.py
@@ -7,7 +7,9 @@ from safe_apps.tests.factories import ProviderFactory
 
 @override_settings(
     CGW_URL="http://127.0.0.1",
+    ALTERNATIVE_CGW_URL="http://alternative.cgw.url",
     CGW_FLUSH_TOKEN="example-token",
+    ALTERNATIVE_CGW_FLUSH_TOKEN="alternative-token",
 )
 class SafeAppHookTestCase(TestCase):
     @responses.activate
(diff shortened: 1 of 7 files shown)


hectorgomezv:
Note: I've left intentionally [a case in the tests](https://github.com/safe-global/safe-config-service/blob/1822ba127f87055a611115e372c935b4d50a9e46/src/chains/tests/test_signals.py#L309) where the new environment variables are not set to verify the service is still working as expected when those are not set.

src/safe_apps/tests/test_signals.py
```diff
@@ -109,10 +195,28 @@ def test_on_provider_create_hook_call(self) -> None:
...
 
-        assert len(responses.calls) == 1
+        assert len(responses.calls) == 2
+        assert isinstance(responses.calls[0], responses.Call)
+        assert responses.calls[0].request.body == b'{"invalidate": "Chains"}'
+        assert responses.calls[0].request.url == "http://127.0.0.1/v2/flush"
+        assert (
+            responses.calls[0].request.headers.get("Authorization")
+            == "Basic example-token"
+        )
```
fmrsabino, line 219: We are asserting the same call twice here.

src/safe_apps/tests/test_signals.py
```diff
@@ -193,10 +363,28 @@ def test_on_tag_create_hook_call(self) -> None:
...
```
fmrsabino, line 387: Same here

fmrsabino (changes requested):

src/safe_apps/tests/test_signals.py
hectorgomezv: Good catch! Fixed in [44d59c7](https://github.com/safe-global/safe-config-service/pull/900/commits/44d59c7c8a3b8845a22ae7d355f56139dd979f5e)

src/safe_apps/tests/test_signals.py
hectorgomezv: Also fixed in [44d59c7](https://github.com/safe-global/safe-config-service/pull/900/commits/44d59c7c8a3b8845a22ae7d355f56139dd979f5e)

fmrsabino (approved):

PR #925, opened 2023-08-15
feat: eip-1559 fixed gas prices

### LINKED ISSUE #924: Fixed gas price settings for EIP-1559
On chain's that support EIP-1559 it is currently not possible to use fixed gas prices for both `maxFeePerGas` and `maxPriorityFeePerGas` as we currently only allow one `weiValue` for fixed gas prices.

**Provide references to the feature you are implementing that requires this change**
For https://github.com/safe-global/safe-wallet-web/issues/2222
we talked with the optimism team and they suggested to use fixed `maxFeePerGas` of 0.1 gwei and `maxPriorityFeePerGas` of 0.0001 gwei. 
This is currently not possible through the chain config of the config service.

I would like to add a new type of gas config `fixed1559`
It requires two new fields `maxFeePerGas` and `maxPriorityFeePerGas` in the `GasPrice` model which are both configured in WEI - analog to the `fixed_wei_value` for the current `fixed` gas config.

**Describe alternatives you've considered**
Alternatively we would have to add custom code to our wallet interfaces and set these gas parameters - overwriting the gas-configs of the chains. 
This would require a new deployment if we wanted to change these settings. 
The team also mentioned that we will need similar settings on other networks such as Base. 
Having this implemented on config service level would enable us to
- Easily change the chain configs if the suggested prices change
- Reuse the solution for multiple networks
- Reuse the solution for the mobile wallet interface to have a consistent experience.

schmanu:
## What this PR resolves

Resolves #924

## How it resolves it

- Adds 2 new fields to `GasPrice` model:
  - `maxFeePerGas` (in Wei)
  - `maxPriorityFeePerGas` (in Wei)
- new serializer for fixed EIP-1559 gas price
- migration


src/chains/models.py
```diff
@@ -147,16 +149,37 @@ class GasPrice(models.Model):
...
-        if (self.fixed_wei_value is not None) == (self.oracle_uri is not None):
+        fixed_wei_defined = self.fixed_wei_value is not None
+        fixed1559_defined = (
+            self.max_fee_per_gas is not None
+            and self.max_priority_fee_per_gas is not None
+        )
+        oracle_defined = self.oracle_uri is not None
+        exactly_one_variant = reduce(
+            xor, [fixed_wei_defined, fixed1559_defined, oracle_defined]
+        )
```
fmrsabino, line 171: This won't work as it is since `True XOR True XOR True` becomes `False XOR True` which becomes `True` and that won't match with "exactly one variant".

fmrsabino (changes requested):
I'd add a test from "view" perspective – if a a `fixed1559` gas price is set, how is it rendered in the response.

src/chains/models.py
fmrsabino: Maybe something like:

```python3
exactly_one_variant = [fixed_wei_defined, fixed1559_defined, oracle_defined].count(True) == 1
```

wdyt?

schmanu: Whoops I missed that one. Your suggestions sounds good. I will change it to that :)

src/chains/models.py
```diff
@@ -147,16 +149,40 @@ class GasPrice(models.Model):
...
-                    "oracle_uri": "An oracle uri or fixed gas price should be provided (but not both)",
-                    "fixed_wei_value": "An oracle uri or fixed gas price should be provided (but not both)",
+                    "oracle_uri": "An oracle uri, fixed gas price or maxFeePerGas and maxPriorityFeePerGas should be \
+                        provided (but not multiple)",
+                    "fixed_wei_value": "An oracle uri, fixed gas price or maxFeePerGas and maxPriorityFeePerGas \
+                         should be provided (but not multiple)",
+                    "max_fee_per_gas": "An oracle uri, fixed gas price or maxFeePerGas and maxPriorityFeePerGas \
+                        should be provided (but not multiple)",
+                    "max_priority_fee_per_gas": "An oracle uri, fixed gas price or maxFeePerGas and \
+                        maxPriorityFeePerGas should be provided (but not multiple)",
```
fmrsabino, line 183: nitpick: the error detail can be extracted to a variable and referenced (seems to be the same for very field).

PR #964, opened 2023-09-26
Add Hook Events for Chains and Safe Apps

fmrsabino:
Reference: https://github.com/safe-global/safe-client-gateway/issues/682

- Refactors the WebHook hook logic for the Safe Client Gateway client.
- A WebHook event is now represented as a `@dataclass` – `HookEvent`. This `HookEvent` can have two types:
  * `CHAIN_UPDATE` – represents chain related updates (more specifically updates to `Chain`, `GasPrice`, `Feature` – chain feature, and `Wallet`).
  * `SAFE_APPS_UPDATE` – represents safe app related updates (more specifically updates to `SafeApp`, `Provider`, `Tag` and `Feature` – safe app feature).
- The new web hooks are under a feature preview and can be enabled by setting the environment variable `FF_HOOK_EVENTS` to `true`.
  * If `FF_HOOK_EVENTS` is set to `false`, the `/v2/flush` endpoint is triggered (as before).

### DIFF
--- a/src/chains/tests/test_signals_ff_hook_events.py
+++ b/src/chains/tests/test_signals_ff_hook_events.py
@@ -47,7 +47,7 @@ class ChainNetworkHookWithFFHookEventsTestCase(TestCase):
                     {"Authorization": "Basic example-token"}
                 ),
                 responses.matchers.json_params_matcher(
-                    {"type": "CHAIN_UPDATE", "chain_id": str(chain_id)}
+                    {"type": "CHAIN_UPDATE", "chainId": str(chain_id)}
                 ),
             ],
         )
@@ -69,12 +69,38 @@ class ChainNetworkHookWithFFHookEventsTestCase(TestCase):
 
     @responses.activate
     def test_on_chain_delete(self) -> None:
-        chain = ChainFactory.create()
+        # Deleting an object sets the primary key to None so we set it in a separate variable
+        chain_id = fake.pyint()
+        chain = ChainFactory.create(id=chain_id)
+        responses.add(
+            responses.POST,
+            "http://127.0.0.1/v1/hooks/events",
+            status=200,
+            match=[
+                responses.matchers.header_matcher(
+                    {"Authorization": "Basic example-token"}
+                ),
+                responses.matchers.json_params_matcher(
+                    {"type": "CHAIN_UPDATE", "chainId": str(chain.id)}
+                ),
+            ],
+        )
 
         chain.delete()
 
         # 2 calls: one for creation and one for deletion
         assert len(responses.calls) == 2
+        assert isinstance(responses.calls[1], responses.Call)
+        assert responses.calls[
+            1
+        ].request.body == f'{{"type": "CHAIN_UPDATE", "chainId": "{chain_id}"}}'.encode(
+            "utf-8"
+        )
+        assert responses.calls[1].request.url == "http://127.0.0.1/v1/hooks/events"
+        assert (
+            responses.calls[1].request.headers.get("Authorization")
+            == "Basic example-token"
+        )
 
     @responses.activate
     def test_on_chain_update(self) -> None:
@@ -269,7 +295,7 @@ class WalletHookTestCase(TestCase):
         assert len(responses.calls) == 0
 
     @responses.activate
-    def test_on_feature_update_with_chain(self) -> None:
+    def test_on_wallet_update_with_chain(self) -> None:
         chain = ChainFactory.create()
         wallet = FeatureFactory.create(key="Test Wallet", chains=(chain,))
 
@@ -313,7 +339,7 @@ class GasPriceHookTestCase(TestCase):
                     {"Authorization": "Basic example-token"}
                 ),
                 responses.matchers.json_params_matcher(
-                    {"type": "CHAIN_UPDATE", "chain_id": str(self.chain.id)}
+                    {"type": "CHAIN_UPDATE", "chainId": str(self.chain.id)}
                 ),
             ],
         )
--- a/src/chains/tests/test_signals_ff_hook_events.py
+++ b/src/chains/tests/test_signals_ff_hook_events.py
@@ -219,6 +219,39 @@ class FeatureHookTestCase(TestCase):
             == "Basic example-token"
         )
 
+    @responses.activate
+    def test_on_feature_update_with_multiple_chains(self) -> None:
+        chain_1 = ChainFactory.create()
+        chain_2 = ChainFactory.create()
+
+        FeatureFactory.create(key="Test Feature", chains=(chain_1, chain_2))
+
+        # 1 call for Chain 1 creation, 1 call for Chain 2 creation, 1 call for feature creation,
+        # 1 call for Chain 1 M2M update, 1 call for Chain 2 M2M update, 1 call for Feature update
+        assert len(responses.calls) == 6
+        assert isinstance(responses.calls[3], responses.Call)
+        assert isinstance(responses.calls[4], responses.Call)
+        assert responses.calls[
+            3
+        ].request.body == f'{{"type": "CHAIN_UPDATE", "chainId": "{chain_2.id}"}}'.encode(
+            "utf-8"
+        )
+        assert responses.calls[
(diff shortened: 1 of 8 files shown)


src/chains/tests/test_signals_ff_hook_events.py
```diff
@@ -0,0 +1,343 @@
...
+            "utf-8"
+        )
+        assert responses.calls[0].request.url == "http://127.0.0.1/v1/hooks/events"
+        assert (
+            responses.calls[0].request.headers.get("Authorization")
+            == "Basic example-token"
+        )
+
+    @responses.activate
+    def test_on_feature_delete(self) -> None:
```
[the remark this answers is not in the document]
fmrsabino: Makes sense 👍 I'll update them.

src/chains/tests/test_signals_ff_hook_events.py
```diff
@@ -0,0 +1,343 @@
...
+    @responses.activate
+    def test_on_chain_update(self) -> None:
+        chain = ChainFactory.create()
+
+        # Not updating using queryset because hooks are not triggered that way
+        chain.currency_name = "Ether"
+        chain.save()
+
+        # 2 calls: one for creation and one for updating
+        assert len(responses.calls) == 2
```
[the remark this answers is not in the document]
fmrsabino: > why do these tests not check the response body like the other ones?

The assertions should be there indeed. Will fix 👍

src/safe_apps/tests/test_signals_ff_hook_events.py
```diff
@@ -131,24 +131,13 @@ def test_on_provider_create_with_safe_app(self) -> None:
 
         # Safe App Creation, Safe App Update
         assert len(responses.calls) == 2
-        assert isinstance(responses.calls[0], responses.Call)
         assert isinstance(responses.calls[1], responses.Call)
-        assert responses.calls[
-            0
-        ].request.body == f'{{"type": "SAFE_APPS_UPDATE", "chainId": "{chain_id}"}}'.encode(
-            "utf-8"
-        )
```
fmrsabino: I've removed these assertions because they are covered by other tests (i.e.: the creation step is already covered by other tests)

src/chains/tests/test_signals_ff_hook_events.py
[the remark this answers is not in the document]
fmrsabino: Changed naming here: [d465bf0](https://github.com/safe-global/safe-config-service/pull/964/commits/d465bf08bcf9c918fa2354a23c992fc9f775f40a)

Also removed the `hook_call` suffix because I think it's not required given the name os the tests.

src/chains/tests/test_signals_ff_hook_events.py
```diff
@@ -0,0 +1,376 @@
...
+        )
+
+    @responses.activate
+    def test_on_chain_delete(self) -> None:
+        chain = ChainFactory.create()
+
+        chain.delete()
+
+        # 2 calls: one for creation and one for deletion
+        assert len(responses.calls) == 2
```
hectorgomezv, line 77: Should we test here the `CHAIN_UPDATE` webhook payload for a deletion event as we do for creation/update?

src/chains/tests/test_signals_ff_hook_events.py
```diff
@@ -0,0 +1,376 @@
...
+        assert responses.calls[
+            3
+        ].request.body == f'{{"type": "CHAIN_UPDATE", "chainId": "{chain.id}"}}'.encode(
+            "utf-8"
+        )
+        assert responses.calls[3].request.url == "http://127.0.0.1/v1/hooks/events"
+        assert (
+            responses.calls[3].request.headers.get("Authorization")
+            == "Basic example-token"
+        )
```
hectorgomezv, line 194: Nitpick: do you think it would be useful to have a test for a feature update where the feature is associated with more than 1 chain? Including a second chain into one of these tests would suffice IMO. Just to test it iterates through chains when CRUDing features. 

This also applies to wallet-related webhooks tests.

src/chains/tests/test_signals_ff_hook_events.py
```diff
@@ -0,0 +1,376 @@
...
+
+        wallet.save()  # create
+        wallet.key = "Test Wallet v2"
+        wallet.save()  # update
+
+        # Updating a wallet with no chains should not trigger any webhook
+        assert len(responses.calls) == 0
+
+    @responses.activate
+    def test_on_feature_update_with_chain(self) -> None:
```
hectorgomezv, line 272: Is this test name accurate? It seems `feature` is used in the name but the test is referred to `wallet`, right?

src/safe_apps/tests/test_signals.py
```diff
@@ -1,11 +1,12 @@
...
 from django.test import TestCase, override_settings
 
-from safe_apps.models import SafeApp, Tag
-from safe_apps.tests.factories import ProviderFactory
+from ..models import SafeApp, Tag
+from ..tests.factories import ProviderFactory
 
 
 @override_settings(
+    FF_HOOK_EVENTS=False,
```
hectorgomezv, line 9: Should we add this `FF_HOOK_EVENTS` setting to the other TestCases also? (`ProviderHookTestCase`, `TagHookTestCase`)? I\ understand it should be `false` anyway because it's the default value fo the setting, but I think it would be nice to add it as you did for the other TestCases, for the sake of consistency.

src/safe_apps/tests/test_signals_ff_hook_events.py
```diff
@@ -0,0 +1,358 @@
...
+
+        tag.delete()  # delete
+
+        assert len(responses.calls) == 0
+
+    @responses.activate
+    def test_on_tag_delete_with_safe_app(self) -> None:
+        chain_id = fake.pyint()
+        safe_app = SafeAppFactory.create(chain_ids=[chain_id])
+        tag = TagFactory.create(safe_apps=(safe_app,))
```
hectorgomezv, line 266: Same as I commented before, what about having a test with more than one item (`safe_app` in this case), to verify iterations?

hectorgomezv (changes requested):

src/safe_apps/tests/test_signals.py
fmrsabino: We should! Thanks for pointing that out. We should not assume a default for the tests in this case 👍

fmrsabino: Addressed in [bfcbbbf](https://github.com/safe-global/safe-config-service/pull/964/commits/bfcbbbfc65b52d790114b3960c1e3dce3b2e30be)

src/chains/tests/test_signals_ff_hook_events.py
fmrsabino: Addressed in [bfcbbbf](https://github.com/safe-global/safe-config-service/pull/964/commits/bfcbbbfc65b52d790114b3960c1e3dce3b2e30be)

src/chains/tests/test_signals_ff_hook_events.py
fmrsabino: Addressed in [bfcbbbf](https://github.com/safe-global/safe-config-service/pull/964/commits/bfcbbbfc65b52d790114b3960c1e3dce3b2e30be)

src/safe_apps/tests/test_signals_ff_hook_events.py
fmrsabino: Done in [e4134f3](https://github.com/safe-global/safe-config-service/pull/964/commits/e4134f37a519ab64f57f25deaa388d49cd15a1f9)

src/chains/tests/test_signals_ff_hook_events.py
fmrsabino: Done in [e4134f3](https://github.com/safe-global/safe-config-service/pull/964/commits/e4134f37a519ab64f57f25deaa388d49cd15a1f9)

hectorgomezv (approved):
Great job!! 🚀

PR #991, opened 2023-12-13
Run container as non-root user

hectorgomezv:
Changes:
- Create, assign permissions, and run the application Docker container using a non-root user.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,16 @@ RUN set ex \
     && rm -rf /var/lib/apt/lists/* \
     && chmod +x /usr/bin/tini
 
-COPY . .
+# Group 'python' (GID 999) and user 'python' (uid 999) are created
+RUN groupadd -g 999 python && useradd -u 999 -r -g python python
+
+# App folder and '/nginx' mount point are created with the new user as owner
+RUN mkdir -p /nginx && chown -R python:python /nginx .
+COPY --chown=python:python . .
+
+# Container is ran by the new user
+# UID:GID is used as Kubernetes requires numeric IDs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#securitycontext-v1-core
+USER 999:999
+
 RUN DEFAULT_FILE_STORAGE=django.core.files.storage.FileSystemStorage python src/manage.py collectstatic --noinput
 ENTRYPOINT ["/usr/bin/tini", "--", "./docker-entrypoint.sh"]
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -27,6 +27,16 @@ RUN set ex \
     && rm -rf /var/lib/apt/lists/* \
     && chmod +x /usr/bin/tini
 
-COPY . .
+# Group 'python' (GID 999) and user 'python' (uid 999) are created
+RUN groupadd -g 999 python && useradd -u 999 -r -g python python
+
+# App folder and '/nginx' mount point are created with the new user as owner
+RUN mkdir -p /nginx && chown -R python:python /nginx .
```
fmrsabino, line 34: You can colapse these two `RUN` into one to reduce the number of layers produced by the image.

```suggestion
# Group 'python' (GID 999) and user 'python' (uid 999) are created
RUN groupadd -g 999 python && \
    useradd -u 999 -r -g python python && \
    # App folder and '/nginx' mount point are created with the new user as owner
    mkdir -p /nginx && chown -R python:python /nginx .
```

Dockerfile
```diff
@@ -27,6 +27,16 @@ RUN set ex \
...
+# Group 'python' (GID 999) and user 'python' (uid 999) are created
+RUN groupadd -g 999 python && \
+    useradd -u 999 -r -g python python && \
+    # App folder and '/nginx' mount point are created with the new user as owner
+    mkdir -p /nginx && chown -R python:python /nginx .
+COPY --chown=python:python . .
+
+# Container is ran by the new user
+# UID:GID is used as Kubernetes requires numeric IDs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#securitycontext-v1-core
+USER 999:999
```
fmrsabino, line 39: Given the context of the PR and the relatively small scope I don't think it's worth it to add something like an [ARG](https://docs.docker.com/engine/reference/builder/#arg) to represent this.

An `ARG` also exposes this as a build-time argument to be consumed. I don't think it should be an `ENV` either 🤔

fmrsabino (approved):
Awesome! 👏

Dockerfile
hectorgomezv: Do you mean storing UID/GID/username into variables? Yeah, we might do that, but I agree it doesn't add a lot of meaning and it doesn't improve readability a lot given the small scope 👍

fmrsabino: Yes it was about `UID/GID` 👍

PR #1007, opened 2023-12-22
Reduce file/folder copying scope

fmrsabino:
Instead of copying the entire directory, copy only the required files:
- `/src` directory which contains the source code of the service
- `/static` directory which contains the required static files
- `docker-entrypoint.sh` which is the script to be used to start up the service.


hectorgomezv (approved):
👏🏻

PR #1020, opened 2024-01-10
Invalidate safe apps for removed chains

hectorgomezv:
Closes https://github.com/safe-global/safe-client-gateway/issues/1005

**Problem:**
When the `Chains` related to a `SafeApp` configuration change, hooks of type `SAFE_APPS_UPDATE` are dispatched for each chain involved. This is true when the `SafeApp` is created, and also for the `Chain` items linked after the `SafeApp` update, but right now no hook is dispatched for the `Chain` items being unlinked by the update operation.

**Changes:**
- Decouples `post_delete` signal from the affected flow, as it remains as it is.
- Change the handling of `post_save` signal to handle `pre_save` signal, to have access to the instance's `Chain` list before the update is made.
- Both the `Chain` items related to the `SafeApp` before and after the update are stored in a `Set` to avoid repetitions. Hooks are dispatched for all the `Chain` items in the set.

### DIFF
--- a/src/safe_apps/signals.py
+++ b/src/safe_apps/signals.py
@@ -3,7 +3,13 @@ from typing import Any
 
 from django.conf import settings
 from django.core.cache import caches
-from django.db.models.signals import m2m_changed, post_delete, post_save, pre_delete
+from django.db.models.signals import (
+    m2m_changed,
+    post_delete,
+    post_save,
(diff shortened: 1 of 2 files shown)


fmrsabino (approved):
Great work! 👏 🚀

PR #1026, opened 2024-01-16
Remove unused FF_HOOK_EVENTS

hectorgomezv:
**Context:**
The Safe Client Gateway currently accepts two ways to invalidate cache registries: 
- Via sending a POST request to the `/v2/flush/` endpoint containing an _invalidation pattern_.
- Via sending a POST request to the `/v1/hooks/events/` endpoint containing an _event payload_.

The first approach was being used by the Config Service exclusively, when it is configured to do so (i.e.: when `FF_HOOK_EVENTS` feature flag is set). This PR removes the possibility of using this first approach in favor of using the `/v1/hooks/events/` endpoint instead.

**Changes:**
- Removes `FF_HOOK_EVENTS` from the service configuration.
- Changes `signals.py` classes for both Chains and Safe Apps, so these classes don't use the `/v2/flush/` endpoint anymore. (And modifies their test counterparts accordingly)
- Removes the `flush` function from the Safe Client Gateway client implementation.

**Out of scope:**
- Remove/rename `CGW_FLUSH_TOKEN` to an alternative name. This would imply changes in other files and some coordination, so it will be tackled separately.


src/chains/tests/test_signals.py
```diff
@@ -1,88 +1,101 @@
...
+    @responses.activate
+    @override_settings(CGW_URL="http://127.0.0.1", CGW_FLUSH_TOKEN=None)
+    def test_no_cgw_call_with_no_token(self) -> None:
+        ChainFactory.create()
+
+        assert len(responses.calls) == 0
+
+
+@override_settings(CGW_URL="http://127.0.0.1", CGW_FLUSH_TOKEN="example-token")
+class ChainNetworkHookWithFFHookEventsTestCase(TestCase):
```
iamacook, line 29: Nit: let's remove "FF" from the name as we are no longer testing that (from my understanding).

iamacook (approved):

fmrsabino (approved):
👏

iamacook (approved):

Issue #1021, opened 2024-01-12
Add chain logos


katspaugh:
## Problem description

The design needs of our web app have evolved, and we want to display chain logos in the chain selector and in various other places in the app. See screenshot at the bottom.

Logos are currently hard-coded in the web app for each chain, so it would be great to have an uplodable field in Django to get them from the config service.

## Requirements
Please extend the chain config with the following fields:

* `chainLogo`
  – Must accept SVGs and PNGs w/o size limitations
  – Must be a required field

## Screenshot for reference
<img width="196" alt="Screenshot 2024-01-12 at 15 38 29" src="https://github.com/safe-global/safe-config-service/assets/381895/8f619016-5331-401d-8d15-f220d7a2d9a7">

fmrsabino:
I've moved the `isTestnet` part to https://github.com/safe-global/safe-config-service/issues/1022. API proposal and requirements to be added.

iamacook:
> For PNGs, file size is more important than dimensions.

There would be no file size limitation.

iamacook:
> My point is, if you want to add any limitation, it should be on the file size, not the dimensions. Otherwise I don't have a preference.

[Pre-existing validation](https://github.com/safe-global/safe-config-service/blob/265d739f39f6857717f32a4d697a9326d43243b9/src/chains/models.py#L38) will be used for consistency, hence wanting to confirm. As you have no preference regarding this, I will continue with the current implementation.

katspaugh:
As stated in the Requirements in the description, the upload field should accept SVG files which is not the case with the current production version.

fmrsabino:
> As stated in the Requirements in the description, the upload field should accept SVG files which is not the case with the current production version.

The API in its current state needs to accommodate mobile platforms which have limited SVG support. SVG support can be considered in future iterations where we might need to provide not only PNGs but SVGs (both returned). This might mean either a new field or a filter on the resource to retrieve different file types.

Due to the scope of this change and since we have other fields which deal with PNG resources this won't be considered in the scope of "adding chain logos", but can of course be considered in the future once there's alignment on the support that we should provide for image resources.

PR #1039, opened 2024-02-05
Add `chainLogoUri` field to `Chain` model

iamacook:
This adds a new `chainLogoUri` `ImageField` to the `Chain` model which returns an absolute URL to the uploaded image, as well as adding/updating associated tests.

The implementation matches that of `currency_logo_uri` as [has been confirmed by the clients](https://github.com/safe-global/safe-config-service/issues/1021#issuecomment-1927023287):

- PNG support (max 512x512px, unrestricted file size)
- Uploaded as `*chain_logo.png`

### DIFF
--- a/src/chains/tests/test_models.py
+++ b/src/chains/tests/test_models.py
@@ -1,7 +1,7 @@
 from decimal import Decimal
 
-import web3
 import factory
+import web3
 from django.core.exceptions import ValidationError
 from django.db import DataError
 from django.test import TestCase, TransactionTestCase
@@ -23,7 +23,9 @@ class ChainLogoTestCase(TestCase):
     def test_chain_logo_upload_path(self) -> None:
         chain = ChainFactory.create()
 
-        self.assertEqual(chain.chain_logo_uri.url, f"/media/chains/{chain.id}/chain_logo.jpg")
+        self.assertEqual(
+            chain.chain_logo_uri.url, f"/media/chains/{chain.id}/chain_logo.jpg"
+        )
 
     def test_image_max_size_validation(self) -> None:
         chain = ChainFactory.create(


src/chains/tests/test_models.py
```diff
@@ -18,6 +18,13 @@ def test_str_method_outputs_name_chain_id(self) -> None:
         )
 
 
+class ChainLogoTestCase(TestCase):
+    def test_chain_logo_upload_path(self) -> None:
+        chain = ChainFactory.create(id=69)
+
+        self.assertEqual(chain.chain_logo_uri.url, "/media/chains/69/chain_logo.jpg")
```
fmrsabino, line 25: You can use the generated `chain.id` (generated by Faker) to validate the path also (to cover random chain ids).

src/chains/tests/test_views.py
```diff
@@ -295,6 +301,29 @@ def test_null_ens_registry_address(self) -> None:
         self.assertEqual(response.json()["ensRegistryAddress"], None)
 
 
+class ChainLogoTests(APITestCase):
```
fmrsabino, line 304: These tests should be moved to `src/chains/tests/test_models.py` as we are not testing the view directly here.

Also extending `TestCase` should be enough for these tests.

fmrsabino (changes requested):

src/chains/tests/test_models.py
iamacook: I adjusted it in 82083887dcd88a0e34c9b70fe269a468686c90e2, along with that of the `currency_logo` for consistency.

src/chains/tests/test_views.py
iamacook: I moved it in 82083887dcd88a0e34c9b70fe269a468686c90e2, along with that of the `currency_logo` for consistency.

fmrsabino (approved):
🚀

PR #1107, opened 2024-04-22
Fix invalid iconUrl on Safe App creation

### LINKED ISSUE #1100: Bad logo URL when creating a new Safe App
**Describe the bug**
When inserting a new Safe App, `None` is added to the logo image URL instead of the `app_id`. 

Re-uploading the image for the Safe App solves the problem.

**To Reproduce**
- Create a new Safe App.
- Check the path for the logo image is not correct (it includes `None` as ID).

**Expected behavior**
A correct Safe App `app_id` is added to the logo path instead of `None`.

**Environment (please complete the following information):**
 - Staging and production.

hectorgomezv:
Closes #1100 

## Summary
This PR fixes a bug in the `SafeApp` creation flow. As `SafeApp` is not persisted to the database yet when `safe_app_icon_path` is executed, it has no access to the `app_id` field (as it depends on the database PK). So this field cannot be used directly as part of the file URL. As a result, before this PR, `None` was being appended to the URL instead of the app ID.

## Changes
- Generate logo URLs by calling `uuid.uuid4()` instead of relying on the Safe App ID.

### DIFF
--- a/src/safe_apps/models.py
+++ b/src/safe_apps/models.py
@@ -1,4 +1,5 @@
 import os
+import uuid
 from enum import Enum
 from typing import IO, Union
 
@@ -17,12 +18,7 @@ _HOSTNAME_VALIDATOR = RegexValidator(
 
 def safe_app_icon_path(instance: "SafeApp", filename: str) -> str:
     _, file_extension = os.path.splitext(filename)
-    if instance.app_id is None:  # new SafeApp, app_id is not available at this point
-        safe_apps = SafeApp.objects.all().order_by("-app_id")
-        app_id = safe_apps[0].app_id + 1 if safe_apps else 1
-        return f"safe_apps/{app_id}/icon{file_extension}"
-    else:  # existing SafeApp, app_id is available
-        return f"safe_apps/{instance.app_id}/icon{file_extension}"
+    return f"safe_apps/{uuid.uuid4()}/icon{file_extension}"
 
 
 def validate_safe_app_icon_size(image: Union[str, IO[bytes]]) -> None:
--- a/src/safe_apps/models.py
+++ b/src/safe_apps/models.py
@@ -17,7 +17,12 @@ _HOSTNAME_VALIDATOR = RegexValidator(
 
 def safe_app_icon_path(instance: "SafeApp", filename: str) -> str:
     _, file_extension = os.path.splitext(filename)
-    return f"safe_apps/{instance.app_id}/icon{file_extension}"
+    if instance.app_id is None:  # new SafeApp, app_id is not available at this point
+        safe_apps = SafeApp.objects.all().order_by("-app_id")
+        app_id = safe_apps[0].app_id + 1 if safe_apps else 1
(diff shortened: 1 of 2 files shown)


src/safe_apps/models.py
```diff
@@ -17,7 +17,12 @@
 
 def safe_app_icon_path(instance: "SafeApp", filename: str) -> str:
     _, file_extension = os.path.splitext(filename)
-    return f"safe_apps/{instance.app_id}/icon{file_extension}"
+    if instance.app_id is None:  # new SafeApp, app_id is not available at this point
+        safe_apps = SafeApp.objects.all().order_by("-app_id")
+        app_id = safe_apps[0].app_id + 1 if safe_apps else 1
+        return f"safe_apps/{app_id}/icon{file_extension}"
+    else:  # existing SafeApp, app_id is available
+        return f"safe_apps/{instance.app_id}/icon{file_extension}"
```
hectorgomezv, line 25: Since `app_id` is not available at this point, this code mimics the logic of assigning a new ID to the `SafeApp`. If for any reason this calculated ID does not match the DB one (deletion race condition for example), it will be eventually fixed when the `SafeApp` gets updated.

An alternative solution would be using a random UUID instead of the `app_id` field, but I find it useful the URLs match the app ID. I'm open to suggestions though.

hectorgomezv: After an internal sync, we've decided to take the UUID approach to compose the logo URL. (Thanks for your feedback @iamacook)

[e16544e](https://github.com/safe-global/safe-config-service/pull/1107/commits/e16544ea2040bfbeb8d643b5b8bc9456a1cc38ab)

src/safe_apps/tests/test_models.py
```diff
@@ -16,9 +16,14 @@ class IconTestCase(TestCase):
     def test_icon_upload_path(self) -> None:
         safe_app = SafeAppFactory.create()
 
-        self.assertEqual(
-            safe_app.icon_url.url, f"/media/safe_apps/{safe_app.app_id}/icon.jpg"
+        path_regex = "|".join(
+            [
+                r"\/media\/safe_apps\/",
+                r"[0-9(a-f|A-F)]{8}-[0-9(a-f|A-F)]{4}-4[0-9(a-f|A-F)]{3}-[89ab][0-9(a-f|A-F)]{3}-[0-9(a-f|A-F)]{12}",
```
iamacook, line 22: This [allows parentheses and pipes](https://gist.github.com/johnelliott/cf77003f72f889abbc3f32785fa3df8d?permalink_comment_id=4376611#gistcomment-4376611) but I don't think we need worry as it's a test.

iamacook (approved):

src/safe_apps/tests/test_models.py
hectorgomezv: Yes, I agree, I think it doesn't need to be super strict. I was thinking in even using `/\w{8}-\w{4}-\w{4}-\w{4}-\w{12}/` which is simpler and also more permissive. But I also think it's OK as this is a test. Thanks!

PR #1111, opened 2024-04-23
Add optional "onlyListed" query param for SafeApps filtering

hectorgomezv:
## Summary
This PR adds an optional query parameter to `/safe-apps/` endpoint to manage the visibility filter.

## Changes
- Adds `onlyListed` query parameter. When it's `false`, both `visible: true` and `visible: false` Safe Apps are returned in the endpoint response. When it's `true` only `visible: true` Safe Apps are returned.

### DIFF
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -28,7 +28,7 @@ BASE_DIR = Path(__file__).resolve().parent.parent
 SECRET_KEY = os.getenv("SECRET_KEY", "random-dev-key")
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
+DEBUG = os.getenv("DEBUG", "true").lower() == "true"
 
 # https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-ALLOWED_HOSTS
 allowed_hosts = os.getenv("DJANGO_ALLOWED_HOSTS", ".localhost,127.0.0.1,[::1]")
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -28,7 +28,7 @@ BASE_DIR = Path(__file__).resolve().parent.parent
 SECRET_KEY = os.getenv("SECRET_KEY", "random-dev-key")
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.getenv("DEBUG", "true").lower() == "true"
+DEBUG = True
 
 # https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-ALLOWED_HOSTS
 allowed_hosts = os.getenv("DJANGO_ALLOWED_HOSTS", ".localhost,127.0.0.1,[::1]")
--- a/src/safe_apps/views.py
+++ b/src/safe_apps/views.py
@@ -1,4 +1,4 @@
-from typing import Any
+from typing import Any, Union
 
 from django.db.models import Q, QuerySet
 from django.utils.decorators import method_decorator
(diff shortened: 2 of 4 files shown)


src/safe_apps/views.py
```diff
@@ -36,12 +40,20 @@ class SafeAppsListView(ListAPIView):  # type: ignore[type-arg]
         type=openapi.TYPE_STRING,
     )
 
+    _swagger_ignore_visibility_param = openapi.Parameter(
+        "ignoreVisibility",
+        openapi.IN_QUERY,
+        description="Ignore visibility flag, including both visible and not-visible Safe Apps",
+        type=openapi.TYPE_STRING,
```
fmrsabino, line 47:
```suggestion
        type=openapi.TYPE_BOOLEAN,
```

src/safe_apps/views.py
```diff
@@ -52,7 +64,13 @@ def get(self, request: Request, *args: Any, **kwargs: Any) -> Response:
         return super().get(request, *args, **kwargs)
 
     def get_queryset(self) -> QuerySet[SafeApp]:
-        queryset = SafeApp.objects.filter(visible=True)
+        ignore_visibility = parse_boolean_query_param(
+            self.request.query_params.get("ignoreVisibility", False)
+        )
+        if ignore_visibility is True:
```
fmrsabino, line 70:
```suggestion
        if ignore_visibility:
```

fmrsabino (changes requested):
I think we need to first clarify the role of the `visibility` flag before pushing another query param that ignores that same flag.

src/config/settings.py
```diff
@@ -28,7 +28,7 @@
 SECRET_KEY = os.getenv("SECRET_KEY", "random-dev-key")
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.getenv("DEBUG", "true").lower() == "true"
+DEBUG = True
```
fmrsabino, line 31: 🤐

src/safe_apps/views.py
```diff
@@ -52,7 +65,13 @@ def get(self, request: Request, *args: Any, **kwargs: Any) -> Response:
         return super().get(request, *args, **kwargs)
 
     def get_queryset(self) -> QuerySet[SafeApp]:
-        queryset = SafeApp.objects.filter(visible=True)
+        only_listed = parse_boolean_query_param(
+            self.request.query_params.get("onlyListed", True)
```
fmrsabino, line 69: I think it might be preferable to have the default as `False`. An in when filtering a collection, if no filter is provided then I'd expect all the items of the collection.

It also makes querying any other Safe App easier. E.g.: If you want to query by an `url` on any app you'd have to set `onlyListed=false` to do that.

src/safe_apps/views.py
```diff
@@ -52,7 +65,13 @@ def get(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+        )
+        if not only_listed:
+            queryset = SafeApp.objects.all()
+        else:
+            queryset = SafeApp.objects.filter(visible=True)
```
fmrsabino, line 74: nitpick: I'd use the positive "form" / expression instead of negating it.

```python
if only_listed:
    queryset = SafeApp.objects.filter(visible=True)
else:
    queryset = SafeApp.objects.all()
```

fmrsabino (changes requested):

src/safe_apps/views.py
hectorgomezv: Done in [501ce8f](https://github.com/safe-global/safe-config-service/pull/1111/commits/501ce8ffbbed691e5c1e3b48424860ba45489a29)

src/safe_apps/views.py
hectorgomezv: I agree, thanks!  Done in [501ce8f](https://github.com/safe-global/safe-config-service/pull/1111/commits/501ce8ffbbed691e5c1e3b48424860ba45489a29)

fmrsabino (approved):
🚀

PR #1139, opened 2024-05-29
Make `iconUrl` of Safe Apps required

### LINKED ISSUE #1138: Make logo fields required

When adding a Safe App (and maybe other entities), if a logo is not uploaded the model is invalid and can cause the interface to crash. The logo should be a required field (where relevant).

Logo fields are required.

iamacook:
Resolves #1138

## Summary

Our clients expect the `iconUrl` of a Safe App to be defined. If no image is present, they crash. As such, this explicitly prevents nullish images as icons, setting a default image and adding a default URL to any entries that may be missing them.

## Changes

- Prevent Django from storing `null` icons
- Do not allow the field to be `blank`
- Add default image
- Add migration for assigning default value and changing the value to non-nullable (as they can't be performed in the same operation)


src/safe_apps/migrations/0013_alter_safeapp_icon_url.py
```diff
@@ -0,0 +1,48 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("safe_apps", "0012_rename_visible_safeapp_listed"),
+    ]
+
+    operations = [
+        # Changing field from nullable to non-nullable is impossible in the same transaction
+        migrations.RunSQL(
+            "SET CONSTRAINTS ALL IMMEDIATE", reverse_sql="SET CONSTRAINTS ALL DEFERRED"
```
Uxio0, line 30: Why do you need this? It shouldn't be required

src/safe_apps/migrations/0013_alter_safeapp_icon_url.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def add_missing_default_icon_url(
+    apps: StateApps, schema_editor: BaseDatabaseSchemaEditor
+) -> None:
+    SafeApp = apps.get_model("safe_apps", "SafeApp")
+    default_icon_url = "safe_apps/icon_url.jpg"
+
+    SafeApp.objects.filter(icon_url__isnull=True).update(icon_url=default_icon_url)
+    SafeApp.objects.filter(icon_url="").update(icon_url=default_icon_url)
```
Uxio0, line 18: `from django.db.models import Q` at the beginning of the file, then:

```suggestion
    SafeApp.objects.filter(Q(icon_url__isnull=True) | Q(icon_url="")).update(icon_url=default_icon_url)
```

src/safe_apps/migrations/0013_alter_safeapp_icon_url.py
iamacook: As discussed, as setting a value and changing the field to non-nullable is otherwise not possible in the same migration, these two operations have been separated into separate migrations in e2c43374aeb23e17d742a5fca7e24cd3f85fa36b.

.gitignore
```diff
@@ -134,4 +134,4 @@ dmypy.json
 data/
 docker-compose.override.yml
 .vscode/
-src/media/
+src/media/chains
```
hectorgomezv, line 137: What's the reasoning behind this change? I think we need to gitignore all the local files in `src/media/safe-apps`. If the intention is to add the default logo (`icon_url.jpg`) to the repository, we could use the `!` prefix to whitelist 🤔

hectorgomezv (changes requested):

hectorgomezv (approved):

Uxio0 (approved):
Great job @iamacook !

PR #1153, opened 2024-06-12
Add Zerion configuration to Chain

hectorgomezv:
## Summary

This PR adds the fields `balances_provider_chain_name` and `balances_provider_enabled` to the `Chain` model. These are intended to hold the prices provider configuration data per chain. This data would be serialized as follows inside the `Chain` response JSON:

```
[...]
"BalancesProvider": {
  "chainName": "ethereum",
  "enabled": true
},
[...]
```

## Changes
- Add `Chain['balances_provider']` field, holding `chainName` and `enabled` properties.
- Adds a migration to add the fields to the `Chain` database table.
- Adds a test for the database migration.
- Modifies the existing tests to adapt them to the changes.


src/chains/tests/test_views.py
```diff
@@ -70,6 +70,10 @@ def test_json_payload_format(self) -> None:
                         "nativeCoin": chain.prices_provider_native_coin,
                         "chainName": chain.prices_provider_chain_name,
                     },
+                    "counterfactualBalancesProvider": {
```
iamacook, line 73: Perhaps we should consider that the balances provider may be used for more than just counterfactual Safe in the future. It is otherwise very feature-specific.

What do you think about having something like:

```
balancesProvider: {
  chainName: string
  counterfactual: boolean
}
```

iamacook: I think it makes more sense to do so after looking at https://github.com/safe-global/safe-client-gateway/pull/1648.

hectorgomezv: Good point, it makes sense to me! I was trying to distinguish it from the other provider (Coingecko) but I agree it should be enough to have a `balances` prefix, as the other one is just for prices.

Will change it, thanks!

hectorgomezv: Done in [3e75238](https://github.com/safe-global/safe-config-service/pull/1153/commits/3e7523879c89a2481100001ee1a4378e9acfbed6)

hectorgomezv:
I've added a field description to clarify the purpose of the `enabled` boolean field: [6b7707d](https://github.com/safe-global/safe-config-service/pull/1153/commits/6b7707d7480adf7897fef3cd03c872bd8e8288ee)

iamacook (approved):

iamacook (approved):

PR #1233, opened 2024-09-19
Feat(Safe Apps): make chains a model with chain_id and name

### LINKED ISSUE #984: Chain ids in Safe Apps
## Problem
Atm we have to enter raw chain ids in Safe Apps entries.
This is very inconvienient and prone to error.

Screenshot 2023-12-06 at 16 57 41

## Request
It would be much better to use the Chain configs for this, like we do for Features.

Screenshot 2023-12-06 at 16 58 28

katspaugh:
## What it solves

Resolves #984

* Add a new model `Chain` with chain_id and name
* Display it in the admin as a mutliselect

Screenshot 2024-09-19 at 11 08 12

## Migrations:
* Pre-populate the Chains table with the chains we support so far
* Migrate the existing data from the `chain_ids` field to `chains`
* Remove the `chain_ids` field

Screenshot 2024-09-19 at 11 07 54


katspaugh:

PR #1291, opened 2024-11-21
Feat: add descriptions to chain features

katspaugh:
Add a simple description field to the Feature model so that feature toggles can be self-documented.


hectorgomezv (approved):
🚀

PR #1339, opened 2025-04-16
feat: add `zk` flag to `Chain` model

### LINKED ISSUE #1337: Add zkSync flag to chain config.

The web requires custom gas config. for zySync-based chains, which is currently hardcoded client-side. In order to make this dynamic, an chain config. flag is required.

**Provide references to the feature you are implementing that requires this change**
Provide at least one of the following:

- Example client-side implementation for Lens - https://github.com/safe-global/safe-wallet-monorepo/pull/5663

A new `zkSync` flag should be added to the chain config.

**Describe alternatives you've considered**

Continual static config. client-side.

**Additional context**

After added, this change will need be propagated to the Client Gateway and eventually consumed client-side.

iamacook:
## Summary

Resolves #1337

This adds a new `zk` flag to the `Chain` model for client-side reference.

## Changes

- Add flag to model, serializer and factory
- Generate/test migration
- Update existing tests


src/chains/migrations/0047_chain_zk.py
```diff
@@ -0,0 +1,31 @@
...
+        ('chains', '0046_alter_chain_create_call_address_and_more'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='chain',
+            name='zk',
+            field=models.BooleanField(default=False),
+        ),
+        migrations.RunSQL(
```
PooyaRaki, line 18: Wouldn’t it be safer to avoid using a migration for this? Ideally, we should avoid hardcoding values in a migration.

iamacook: It was to avoid having to manually update the values after release but I removed it in af3a32ef606e21c6d37fec3506ca618a6d6f38bb. We will need to remember to update it.

PooyaRaki (approved):

PR #1519, opened 2026-04-07
fix: broadcast CHAIN_UPDATE hook to all registered services

tmjssz:
## What it solves

When Chain, GasPrice, or Wallet models are updated, the CHAIN_UPDATE webhook fires with `service=None`, making it impossible for downstream consumers to perform service-aware cache invalidation. Only Feature-related signals included service keys.

## How this PR fixes it

`ChainUpdateWebhookService.notify()` now queries all registered `Service` records when called without explicit `service_keys`, firing one hook per (chain, service) pair. If no services exist in the DB, it falls back to the current behavior (`service=None`).

A subtle but important fix: Feature signal callers previously coerced empty service lists to `None` via `or None`. Under the new semantics `None` means "broadcast to all services", so this would have caused Features with zero associated services to accidentally notify every service. The `or None` coercion is removed — these callers now pass `[]` to mean "no services".

## How to test it

1. Create two `Service` records (e.g., `CGW`, `WALLET_WEB`)
2. Update a `Chain` model field and observe that two webhook calls are made — one per service key
3. Update a `GasPrice` or `Wallet` and verify the same per-service behavior
4. Delete all `Service` records and update a chain — verify a single webhook fires with no `service` field (fallback)

### DIFF
--- a/src/chains/services.py
+++ b/src/chains/services.py
@@ -1,6 +1,8 @@
+# SPDX-License-Identifier: FSL-1.1-MIT
 import logging
 from collections.abc import Iterable
 
+from chains.models import Service
 from clients.safe_client_gateway import HookEvent, hook_event
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


tmjssz:
@claude review

tmjssz:
@claude review again pls. I addressed your concerns:

692c0bd — Updated notify() to broadcast to all Service records when service_keys=None. Removed or None coercion in signals.py so Feature signals pass [] instead of None. Added tests.
13ac2ce — Added test for no-services fallback (fires hook with service=None when no services exist).
04e36a2 — Added delete-with-services tests for Wallet and GasPrice hooks.
dafbb41 — Added inline comment in signals.py explaining [] vs None semantics.

src/chains/services.py
```diff
@@ -12,6 +14,25 @@ def notify(
...
+        """Send CHAIN_UPDATE hook events for the given chains.
+
+        Args:
+            chain_ids: Chain IDs to notify about.
+            service_keys: Controls which services are notified:
+                - None: query all registered Service records (broadcast).
+                - []: no services — fire one hook per chain with service=None.
+                - ["WALLET_WEB", ...]: fire one hook per (chain, service_key) pair.
+        """
+        if service_keys is None:
```
LucieFaire, line 26: qs: can it now be None?

LucieFaire (approved):

src/chains/services.py
PooyaRaki: cc @tmjssz

tmjssz: Yes, the callers in [signals.py](https://github.com/safe-global/safe-config-service/blob/fix%2Fchain-update-hook-broadcast-all-services/src/chains/signals.py) that handle chain/wallet/gas-price changes don't pass service_keys at all, they call notify(chain_ids) without the second argument, so it defaults to None. That None is what triggers the broadcast path: query all registered services from the DB (see the comment in line 22).

PR #1530, opened 2026-04-14
fix: Revert python version back to a stable one

LucieFaire:
## Summary
Revert the dependabot change to python version and use the old stable one, so the build doesnt break (due to missing deps)


LucieFaire:

LucieFaire:
recheck

PooyaRaki (approved):

PR #1557, opened 2026-05-20
Add configurable chain token list

moisses89:
PLA-1389
## Description
This PR adds support for managing gas tokens across chains, including admin configuration and API exposure.
  - Added new gas token data model and persistence layer
  - Added admin capabilities to manage and enable tokens by chain
  - Added endpoint to expose configured gas tokens per chain
  - Improved related API/documentation compatibility
  - Added test coverage for model, admin, and API behavior


src/chains/models.py
```diff
@@ -268,6 +268,24 @@ def __str__(self) -> str:
         return f"Wallet: {self.key}"
 
 
+class GasToken(models.Model):
+    class Meta:
+        unique_together = ("address", "symbol")
```
falvaradorodriguez: unique_together is redundant, address already has unique=True, which makes any (address, symbol) pair automatically unique. The composite constraint is never independently triggered and just creates a second B-tree index on the same data.

src/chains/serializers.py
```diff
@@ -173,6 +173,17 @@ def to_representation(self, instance: Wallet) -> str:  # type: ignore[override]
...
 
+class GasTokenSerializer(serializers.ModelSerializer[GasToken]):
+    address = EthereumAddressField()
+    enabled_chains: serializers.PrimaryKeyRelatedField[Any] = serializers.PrimaryKeyRelatedField(
+        source="chains", many=True, read_only=True
+    )
+
+    class Meta:
+        model = GasToken
+        fields = ["address", "enabled_chains"]
```
falvaradorodriguez: symbol is missing from the serializer response. Is it necessary to include it in the model?

src/config/urls.py
```diff
@@ -17,6 +19,7 @@
     path("about/", include("about.urls", namespace="about")),
     path("safe-apps/", include("safe_apps.urls", namespace="safe-apps")),
     path("chains/", include("chains.urls", namespace="chains")),
+    path("gas-tokens/", GasTokensListView.as_view(), name="gas-tokens-list"),
```
falvaradorodriguez: GasToken is registered as a top-level resource (/v1/gas-tokens/) rather than a sub-resource of a chain, which signals it deserves its own gas_tokens app with its own models.py, views.py, and urls.py.  WDYT?

falvaradorodriguez (changes requested):

src/chains/serializers.py
moisses89: No, the body response is detailed in the task PLA-1389
I added symbol just to simplify the admin view, so is easier search by USDC for example for future updates.

falvaradorodriguez (dismissed):
LGTM!

katspaugh:
@claude review

src/chains/tests/test_models.py
```diff
@@ -407,6 +408,38 @@ def test_str_method_outputs_name_and_key(self) -> None:
...
+        token = GasTokenFactory.create()
+
+        self.assertEqual(str(token), f"GasToken: {token.symbol} ({token.address})")
+
+    def test_address_and_symbol_combination_is_unique(self) -> None:
+        token = GasTokenFactory.create()
+
+        with self.assertRaises(Exception):
+            GasTokenFactory.create(address=token.address, symbol=token.symbol)
+
```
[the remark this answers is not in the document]
moisses89: address already has unique=True, which makes any (address, symbol) pair automatically unique.

src/chains/admin.py
```diff
@@ -136,6 +137,63 @@ class ServiceAdmin(admin.ModelAdmin[Service]):
...
+        ),
+    )
+    actions = ["enable_for_all_chains"]
+
+    @admin.display(description="Enabled chains")
+    def enabled_chains(self, obj: GasToken) -> str:
+        names = list(obj.chains.values_list("name", flat=True).order_by("name"))
+        if not names:
+            return mark_safe("<span style='color:#999'>Disabled</span>")
+        return ", ".join(names)
```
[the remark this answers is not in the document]
moisses89: Done

src/chains/views.py
```diff
@@ -55,6 +55,14 @@ def get(self, request: Request, *args: Any, **kwargs: Any) -> Response:
         return super().get(request, *args, **kwargs)
 
 
+class GasTokensListView(ListAPIView[GasToken]):
+    serializer_class = GasTokenSerializer
+    pagination_class = ChainsPagination
+
+    def get_queryset(self) -> QuerySet[GasToken]:
+        return GasToken.objects.filter(chains__id=self.kwargs["pk"])
```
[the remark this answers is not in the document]
moisses89: Done https://github.com/safe-global/safe-config-service/pull/1557/changes/70c768ad625570e7a88f39b8208ea48038510862

moisses89:
@katspaugh Claude comments were addressed.

diegopazosrego:
@claude review

moisses89:
> **Claude finished @diegopazosrego's task in 1m 47s** —— [View job](https://github.com/safe-global/safe-config-service/actions/runs/26284705737)
> 
> ### Claude PR Review
> * [x]  Read repo & PR context
> * [x]  Inspect changed files
> * [x]  Compile review feedback
> * [x]  Post review
> 
> The PR is in good shape — the prior round of feedback has been almost entirely addressed: queryset is now ordered, 404 is returned for unknown/hidden chains, `mark_safe` was swapped to `format_html`, the N+1 on `enabled_chains` is fixed via `prefetch_related`, and the chain ids are materialized once in the admin action. Below are the few items that are still open plus a couple of small new observations.
> 
> ### Still open from prior review
> #### 1. `GasTokenSerializer` is missing `ref_name`
> `src/chains/serializers.py:176-181`
> 
> ```python
> class GasTokenSerializer(serializers.ModelSerializer[GasToken]):
>     address = EthereumAddressField()
> 
>     class Meta:
>         model = GasToken
>         fields = ["address", "symbol"]
> ```
> 
> Every other `ModelSerializer` in this file sets `Meta.ref_name` (e.g. `GasPriceSerializer`, `FeatureSerializer`, `WalletSerializer`) so `drf-yasg` produces stable, unambiguous schema names. Suggest:
> 
> ```python
> class Meta:
>     model = GasToken
>     fields = ["address", "symbol"]
>     ref_name = "chains.serializers.GasTokenSerializer"
> ```
> 
> [Fix this →](https://claude.ai/code?q=In%20src%2Fchains%2Fserializers.py%20GasTokenSerializer.Meta%2C%20add%20%60ref_name%20%3D%20%22chains.serializers.GasTokenSerializer%22%60%20so%20drf-yasg%20uses%20a%20stable%20schema%20name%2C%20matching%20the%20pattern%20of%20the%20other%20ModelSerializers%20in%20this%20file.%20Branch%3A%20add_configurable_chain_token_list&repo=safe-global/safe-config-service)
> 
> #### 2. Trailing-comma style on `GasToken.address`
> `src/chains/models.py:277-279`
> 
> ```python
> address = EthereumAddressBinaryField(
>     unique=True, help_text="Token contract address.",
> )
> ```
> 
> The multi-line + trailing comma for two kwargs is inconsistent with similar two-kwarg fields elsewhere in the file (e.g. `Wallet.key`, `Wallet.enable_by_default` are written inline or without the trailing comma). Pure style — drop the comma or inline.
> 
> ### New / smaller observations
> #### 3. `enable_for_all_chains` issues an extra `SELECT COUNT(*)` after the loop
> `src/chains/admin.py:188-197`
> 
> ```python
> def enable_for_all_chains(
>     self, request: HttpRequest, queryset: QuerySet[GasToken]
> ) -> None:
>     all_chain_ids = list(Chain.objects.values_list("pk", flat=True))
>     for gas_token in queryset:
>         gas_token.chains.set(all_chain_ids)
>     self.message_user(
>         request, f"Enabled {queryset.count()} gas token(s) for all chains."
>     )
> ```
> 
> `queryset.count()` after iterating runs an extra `SELECT COUNT(*)`. Since you've already iterated `queryset`, you can capture the count as you go:
> 
> ```python
> all_chain_ids = list(Chain.objects.values_list("pk", flat=True))
> tokens = list(queryset)
> for gas_token in tokens:
>     gas_token.chains.set(all_chain_ids)
> self.message_user(
>     request, f"Enabled {len(tokens)} gas token(s) for all chains."
> )
> ```
> 
> Also worth considering wrapping the loop in `transaction.atomic()` so a partial failure doesn't leave half the selected tokens reassigned — not strictly required since each `set()` is atomic on its own, but it makes the action atomic as a whole, which matches the user-visible "enable for all chains" intent.
> 
> #### 4. `GasTokenChainListFilter` exposes hidden chains
> `src/chains/admin.py:140-153`
> 
> ```python
> def lookups(self, ...):
>     return list(Chain.objects.values_list("id", "name").order_by("name"))
> ```
> 
> Admin-only, so probably fine — but worth noting that hidden chains will appear in the filter dropdown. If the intent is "don't surface hidden chains anywhere", you'd want `Chain.objects.filter(hidden=False)`. Not a blocker.
> 
> #### 5. Test name still mismatches the actual constraint
> `src/chains/tests/test_models.py:417-421`
> 
> ```python
> def test_address_and_symbol_combination_is_unique(self) -> None:
>     token = GasTokenFactory.create()
>     with self.assertRaises(Exception):
>         GasTokenFactory.create(address=token.address, symbol=token.symbol)
> ```
> 
> The model only enforces uniqueness on `address` (`unique_together` was removed). The test passes because address collides, not because the `(address, symbol)` pair collides — they're effectively the same check now. Renaming to `test_address_is_unique` (and dropping the redundant `symbol=` kwarg) would make intent clearer. Also `assertRaises(Exception)` is broad — `IntegrityError` would be more precise.
> 
> #### 6. Pagination ordering isn't covered by a test
> `src/chains/tests/test_views.py:781-861`
> 
> You've added good coverage for empty/cross-chain/multi-chain/404 cases. Now that `order_by("symbol", "id")` is in place, one small test that creates a few tokens with deliberately unsorted symbols and asserts they come back ordered would lock in the contract. Optional.
> 
> ### Nits
> * Inside `GasTokenAdmin.enabled_chains`, `sorted(chain.name for chain in obj.chains.all())` correctly uses the prefetched relation (good — no extra queries per row).
> * The decision to keep `symbol` on the model purely for admin search ergonomics is sensible; a one-line `# kept for admin search; not part of the public API` on the field would encode that decision in the source so it survives the next refactor (per @falvaradorodriguez's earlier discussion thread).
> * v2 deliberately doesn't expose gas tokens — fine, since gas tokens are chain-scoped and v2 is service-aware.
> 
> Overall: only #1 (ref_name) is meaningful before merge; everything else is polish. Looks ready.
> 
> • • Branch: [`add_configurable_chain_token_list`](https://github.com/safe-global/safe-config-service/tree/add_configurable_chain_token_list)

1 - Done b5965c91782e7271670c4f71be576d6c1d48e35d
2 - Done b5965c91782e7271670c4f71be576d6c1d48e35d
3 - Done b5965c91782e7271670c4f71be576d6c1d48e35d
4 - It is expected, we want to edit chains before unhide them 
5 - Done, test removed b5965c91782e7271670c4f71be576d6c1d48e35d
6 - it is not necessary

moisses89:
@claude review

falvaradorodriguez (approved):

LucieFaire (approved):

PooyaRaki (approved):

PR #1558, opened 2026-05-20
feat: Use composite CLA action from github-reusable-workflows

diegopazosrego:
Migrates the CLA Assistant workflow to use the composite action from `github-reusable-workflows ` instead of defining the steps inline. The calling workflow checks out the private repo using REUSABLE_WORKFLOWS_TOKEN and references the action by local path

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -6,6 +6,8 @@ on:
   pull_request_target:
     types: [opened, closed, synchronize]
 
+permissions: {}
+
 jobs:
   CLAssistant:
     runs-on: ubuntu-latest
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -9,31 +9,21 @@ on:
 jobs:
   CLAssistant:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: write
     steps:
+      - name: Checkout reusable workflows
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/cla.yml
luarx, line 8: I would add here `permissions: {}`
So by default all the jobs has no permissions (https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax?utm_source=chatgpt.com#permissions) as we do in other repos like  [this](https://github.com/safe-global/safe-transaction-service/pull/2884/changes)

diegopazosrego: Fixed on https://github.com/safe-global/safe-config-service/pull/1558/changes/029a2dbfe40b31398e7595ae3df253086a327f1d @luarx

luarx (dismissed):

.github/workflows/cla.yml
```diff
@@ -6,34 +6,26 @@ on:
...
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: write
     steps:
+      - name: Checkout reusable workflows
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          repository: safe-global/github-reusable-workflows
```
compojoom, line 22: I think that we should pin a commit here. We should not go with the head of the repository.

diegopazosrego: done @compojoom https://github.com/safe-global/safe-config-service/pull/1558/changes/fd529efa9a4f65c32a32834a2e84240ae5251377

compojoom (dismissed):

PR #1561, opened 2026-05-22
feat: Use composite Claude code review action

diegopazosrego:
## Summary

  - Migrates the Claude code review workflow to use a composite action from
  `safe-global/github-reusable-workflows`
  - Replaces the broken reusable workflow approach (public repos cannot call private reusable workflows)
  - Uses the same pattern as the CLA action: explicit checkout with `REUSABLE_WORKFLOWS_TOKEN` + composite action
  call
  - Pinned to `ed82e15` (v2.4.16)

  ## Changes

  - `.github/workflows/claude-code-review.yml` now checks out `github-reusable-workflows` and calls
  `actions/claude-code-review`
  - Trigger events and permissions remain in this repo
  - Behavior is identical to the previous setup

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -13,11 +13,21 @@ permissions: {}
 
 jobs:
   claude-review:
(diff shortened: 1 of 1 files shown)


diegopazosrego:

PR #1555, opened 2026-05-19
Add enum relayerType

akshay-ap:
Related to: https://linear.app/safe-global/issue/PLA-1267/cgw-refactor-cgw-env-vars-for-relayer-selection-logic

This pull request introduces support for specifying a relayer strategy for chain.

**Model and Database Updates:**

* Added a new `relayer_type` field to the `Chain` model, allowing selection from several relayer strategies (GTF, Relay Fee, Daily Limit, No Fee Campaign) or leaving it empty for no relayer. This includes a new `RelayerType` enum for valid choices. 

**API and Serialization:**

* Updated the `ChainSerializer` to include the new `relayer_type` field, making it available in API responses.

**Testing and Factories:**

* Modified the chain factory to randomly assign a `relayer_type` for testing purposes.
* Updated tests to verify that the `relayer_type` is present in the API payloads.

PR #1584, opened 2026-06-08
feat(relayer): add sponsorship fields and validation to Chain model

akshay-ap:
Related to: https://linear.app/safe-global/issue/PLA-1267/cgw-refactor-cgw-env-vars-for-relayer-selection-logic

This PR is an extension to a recently merged PR: https://github.com/safe-global/safe-config-service/pull/1555
During discussions with the team, we realized that just having a enum related to relayer type is not enough in config service and we need more settings related to relayer. This PR introduces additional relayer settings and all relayer settings are now grouped together in key "relayer" in API response.

Add 3 granular controls related relaying:
- Sponsoring Safe creation
- Sponsoring Safe transactions
- Tenderly simulation before relaying a transaction

These fields are safe to add to config service and are not secrets in CGW. These new fields will replace env var from CGW and later used in Wallet as needed. Though this is a breaking API change from previous release, it is safe to merge because we are not using `relayerType` in production yet.


akshay-ap:
@claude review

yagopv (approved):

yagopv:
LGTM! 🚀

LucieFaire (approved):

PR #1586, opened 2026-06-09
feat: Order chain gas tokens by priority

falvaradorodriguez:
- Related with [PLA-1552](https://linear.app/safe-global/issue/PLA-1552/fee-engine-consume-gas-token-list-ordering-from-config-service). 

Adds manual priority/ordering to the per-chain gas tokens (allowed gas tokens) list so consumers (Fee Engine / CGW) can read the ordered list and honour the ordering as selection priority.
  
The gas-token list already existed in the config service, but there was no way to express ordering. That ordering *is* the selection priority the Fee Engine needs. This adds the missing field and exposes the list already sorted.


LucieFaire:
what is the base branch: `safe-global:main`? I cannot see this PR in the commit history

falvaradorodriguez:
> what is the base branch: `safe-global:main`? I cannot see this PR in the commit history

For some reason, I don’t have permissions for this repository, so I have to submit the pull request from a fork as an external contributor.

PR #1516, opened 2026-03-25
refactor: Remove unused contract address fields from chain config

Zalens:
## Summary
- Removes 9 unused contract address fields (`safeSingletonAddress`, `safeProxyFactoryAddress`, `multiSendAddress`, `multiSendCallOnlyAddress`, `fallbackHandlerAddress`, `signMessageLibAddress`, `createCallAddress`, `simulateTxAccessorAddress`, `safeWebAuthnSignerFactoryAddress`) from the Chain model and API response
- These fields were added in July 2024 but never populated — they appear empty on both staging and production
- Removes the `ContractAddressesSerializer`, model fields, test factory assignments, and test expectations
- Includes a migration to drop the columns from the database

Closes PLA-1142

## Test plan
- [x] All 235 existing tests pass
- [ ] Verify staging API response no longer includes `contractAddresses`
- [ ] Confirm no downstream clients depend on these fields

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1595, opened 2026-06-22
fix: renumber dupe migration 0058 to 0059

LucieFaire:
## What

Resolves a conflicting-migrations error in the `chains` app: two leaf nodes both
numbered `0058` (`0058_gastoken_priority` and `0058_remove_contract_address_fields`),
each branched off `0057_chain_relayer_sponsoring` in separate PRs (#1587, #1516)
that landed on `main` independently.

CommandError: Conflicting migrations detected; multiple leaf nodes in the
migration graph: (0058_gastoken_priority, 0058_remove_contract_address_fields in chains).

## How

Renumbered `0058_remove_contract_address_fields` → `0059_remove_contract_address_fields`
and repointed its dependency from `0057_chain_relayer_sponsoring` to
`0058_gastoken_priority`, linearizing the graph:

0057 → 0058_gastoken_priority → 0059_remove_contract_address_fields

This follows the same pattern as `5660bb9` ("renumber duplicate 0055 migration to 0056"),
keeping history linear instead of adding a no-op `--merge` migration.

The two operations are independent (one adds `gastoken.priority`, the other removes
unused `chain.*_address` fields), so ordering them linearly is behaviorally a no-op.

`uv run python src/manage.py makemigrations --check --dry-run` now reports
**No changes detected**.

### DIFF
--- a/src/chains/migrations/0059_remove_contract_address_fields.py
+++ b/src/chains/migrations/0059_remove_contract_address_fields.py
@@ -7,7 +7,7 @@ from django.db import migrations
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('chains', '0057_chain_relayer_sponsoring'),
+        ('chains', '0058_gastoken_priority'),
     ]
 
     operations = [


vmosharova (approved):
👍🏻

PR #1600, opened 2026-06-26
feat: Add Datadog APM instrumentation via ddtrace SDK

diegopazosrego:
## Summary

Adds opt-in Datadog APM instrumentation via the ddtrace **SDK** (v4.10.5). Gated by `DD_APM_ENABLED`: when absent or `"false"` (the default), ddtrace is **never imported**, zero overhead and no behaviour change for existing deployments. Requires a companion chart PR in `legacy-safe-argocd-mgmt` to activate (see below).

---

## Background — SDK vs Admission Controller

The Helm chart already ships `apm.enabled`, which uses the Cluster Agent Admission Controller (same mechanism as `safe-client-gateway-nest` for its JS lib). That path is intentionally **not** used here for two reasons:

**1. Python 3.14 + Django 6.0 compatibility**

The chart currently pins `apm.pythonLib: "v4.3.0"`. That version predates official support for Python 3.14 and Django 6.0. ddtrace 4.10.5 (released 2026-06-17) is the first release with PyPI classifiers and a CI matrix that covers both:

| Axis | ddtrace 4.3.0 | ddtrace 4.10.5 |
|---|---|---|
| Python 3.14 | ✗ | ✅ |
| Django 6.0 | ✗ | ✅ |
| psycopg 3.x | partial | ✅ |

**2. No double-initialisation risk**

The Admission Controller wraps the process with `ddtrace-run`, which calls `patch_all()` before our code runs. If both `apm.enabled: true` and `DD_APM_ENABLED: true` were active simultaneously, ddtrace would initialise twice with different versions. The one injected by the Admission Controller (from the cluster's python-lib image) and the one
pinned in `pyproject.toml`. The SDK approach with `apm.enabled: false`  avoids this entirely.

The Admission Controller remains `enabled: true` at cluster level and continues to serve other services (e.g. CGW). This PR simply does not opt this service's pods into it.

---

### `src/apm.py` *(new)*

Top-level module at `src/` , * not inside `src/config/` *so it can be  imported by both `wsgi.py` and clients/safe_client_gateway.py` without introducing an upward dependency from `clients` into `config`.

- `ENABLED` is evaluated once at import time from `DD_APM_ENABLED`.   When `False`, `ddtrace` is never touched anywhere in the process.
- `setup_tracing()` calls `patch(django=True, requests=True,   psycopg=True)`. Patches only the three integrations actually used   rather than `patch_all()`.
- `trace()` returns a real ddtrace span context manager when APM is on,   or `contextlib.nullcontext()` when off — callers use the same `with apm.trace(...) as span:` pattern in both cases.

### `src/config/wsgi.py`

`setup_tracing()` is called **before** `get_wsgi_application()`. This ordering is mandatory: `patch()` must wrap Django's internal module imports before they happen, otherwise the middleware stack and ORM query hooks are missed.

### `src/clients/safe_client_gateway.py`

`hook_event()` — the single exit point for all webhook notifications to CGW regardless of which Django signal triggered them is wrapped in a manual span `cgw.hook_event`.

Tags on the span:
- `cgw.chain_id` :  chain identifier from the hook payload
- `cgw.service` : service name when present
- `error: True` + `error.message` :  populated in the `except` block,   surfaces failures (missing `CGW_URL`, HTTP errors, etc.) in the  Datadog APM error tracking view as first-class events rather than   buried log lines

The `requests` integration patch traces the outbound HTTP POST to CGW as a child span automatically, giving end-to-end visibility from the Django request that triggered the signal down to the outbound call.

### `pyproject.toml`

Pins `ddtrace==4.10.5`. Exact pin is intentional and consistent with the rest of the dependency list; avoids silent upgrades that could break Python 3.14 or Django 6.0 compatibility.

### `.dev.env`

`DD_APM_ENABLED=false` (default — zero overhead in local dev). All other Datadog variables are present but commented out, with inline documentation for each.

---

## What gets traced when enabled

| Span | Source | Key tags |
|---|---|---|
| `django.request` | auto (`patch(django=True)`) | `http.method`, `http.url`, `http.status_code`, view name |
| `psycopg.query` (child) | auto (`patch(psycopg=True)`) | `db.statement`, `db.name` |
| `requests.request` (child) | auto (`patch(requests=True)`) | `http.url`, `http.method`, `http.status_code` |
| `cgw.hook_event` (manual) | `apm.trace(...)` in `hook_event()` | `cgw.chain_id`, `cgw.service`, `error`, `error.message` |

---

## Companion chart PR required — `legacy-safe-argocd-mgmt`

This PR alone does nothing at runtime  `DD_APM_ENABLED` defaults to `false`. The following changes activate it and should go in a separate PR to [legacy-safe-argocd-mgmt](https://github.com/safe-global/legacy-safe-argocd-mgmt)

### 1. [charts/safe-config-service-chart/values.yaml](https://github.com/safe-global/legacy-safe-argocd-mgmt/blob/main/charts/safe-config-service-chart/values.yaml)

Add `apm.sdkEnabled` (new key, distinct from `apm.enabled` which 
controls the Admission Controller — do **not** touch `apm.enabled`):

```yaml
apm:
  enabled: false       # existing — Admission Controller. Must stay false.
  sdkEnabled: false    # new — SDK gate. Override to true in ArgoCD values.
  pythonLib: "v4.3.0"  # existing — unused while apm.enabled is false
  debug: false         # existing
```
### 2.  [charts/safe-config-service-chart/templates/web-deployment.yaml](https://github.com/safe-global/legacy-safe-argocd-mgmt/blob/main/charts/safe-config-service-chart/templates/web-deployment.yaml)

Add after the existing `{{- if .Values.apm.enabled }}` block, inside the web container's env: section:
```
{{- if .Values.apm.sdkEnabled }}
- name: DD_APM_ENABLED
  value: "true"
- name: DD_AGENT_HOST          # DaemonSet agent on the same node (TCP 8126)
  valueFrom:
    fieldRef:
      fieldPath: status.hostIP
- name: DD_ENV
  valueFrom:
    fieldRef:
      fieldPath: metadata.labels['tags.datadoghq.com/env']
- name: DD_SERVICE
  valueFrom:
    fieldRef:
      fieldPath: metadata.labels['tags.datadoghq.com/service']
- name: DD_VERSION
  valueFrom:
    fieldRef:
      fieldPath: metadata.labels['tags.datadoghq.com/version']
- name: DD_LOGS_INJECTION
  value: "true"
- name: DD_TRACE_SAMPLE_RATE
  value: "1"
{{- end }}
```
`DD_ENV/SERVICE/VERSION` read from the pod's own Unified Service Tagging labels already set unconditionally by the chart. No new label definitions needed.

`DD_AGENT_HOST` uses status.hostIP because the Datadog DaemonSet has portEnabled: true at port 8126, each node's agent accepts TCP traces from pods on that same node. The Admission Controller's configMode: service routing is intentionally bypassed here.

### 3. ArgoCD app / staging 

In  [devstaging/argocd-teams-apps/argocd-project-interfaces-backend/safe-config-service.yaml](https://github.com/safe-global/legacy-safe-argocd-mgmt/blob/main/devstaging/argocd-teams-apps/argocd-project-interfaces-backend/safe-config-service.yaml) , inside `helm.valuesObject`:
```
apm:
  sdkEnabled: true
```
### 4. ArgoCD app — production

Same addition in  [production/argocd-teams-apps/argocd-project-interfaces-backend/safe-config-service.yaml](https://github.com/safe-global/legacy-safe-argocd-mgmt/blob/main/production/argocd-teams-apps/argocd-project-interfaces-backend/safe-config-service.yaml)

---
## Local verification

Tested with Python 3.14.4, Gunicorn, PostgreSQL (Podman), and a Datadog agent container (site: datadoghq.eu):

- 271 tests pass with DD_APM_ENABLED=false and no regressions
```
271 passed, 194 warnings, 102 subtests passed in 14.36s 
```
- APM disabled: no ddtrace lines in gunicorn output, zero import overhead confirmed
```
DD_APM_ENABLED=false POSTGRES_HOST=localhost .venv/bin/python src/manage.py check                                   
System check identified no issues (0 silenced).
```
- APM enabled, no agent: all three integrations patched (django, requests, psycopg confirmed in startup logs), app responds  normally, background connection warnings only  no crash
- APM enabled, live agent: django.request spans visible with child psycopg.query spans; cgw.hook_event span present witherror=True and error.message populated (expected CGW_URL not configured in local dev)

--- 
## Images from Traces and Spans
image

image

image

image

## Datadog Links:
[Traces](http://app.datadoghq.eu/apm/traces?agg_m_a=count&agg_m_source_a=base&agg_t_a=count&cols=service%2Cresource_name%2C%40duration%2C%40http.method%2C%40http.status_code%2C%40_span.count%2C%40_duration.by_service&colsTraces=service%2C%40duration%2C%40_span.count%2Cresource_name%2C%40_duration.by_service&fromUser=false&graphType=waterfall&historicalData=true&messageDisplay=inline&query_a=env%3Alocal&query_translation_version=v0&shouldShowLegend=true&sort=desc&spanType=all&spanViewType=databases&storage=hot&target-span=a&traceQuery=a&view=traces&viz=stream&start=1781887473285&end=1782492273285&paused=false)

[Spans](https://app.datadoghq.eu/apm/traces?query=env%3Alocal&agg_m=count&agg_m_source=base&agg_t=count&cols=service%2Cresource_name%2C%40duration%2C%40http.method%2C%40http.status_code%2C%40_span.count%2C%40_duration.by_service&colsTraces=service%2C%40duration%2C%40_span.count%2Cresource_name%2C%40_duration.by_service&fromUser=false&graphType=waterfall&historicalData=true&messageDisplay=inline&shouldShowLegend=true&sort=desc&spanType=all&spanViewType=databases&storage=hot&target-span=a&tq_query_translation_version=v0&traceQuery=a&view=spans&viz=stream&start=1781887501625&end=1782492301625&paused=false)


.dev.env
```diff
@@ -123,3 +123,28 @@ CSRF_TRUSTED_ORIGINS="http://localhost:${NGINX_HOST_PORT}"
...
+
+# Correlación trazas-logs: inyecta trace_id/span_id en cada log record
+#DD_LOGS_INJECTION=true
+
+# Propagación de contexto W3C Trace Context + formato nativo Datadog
+#DD_TRACE_PROPAGATION_STYLE=datadog,tracecontext
+
+# Toggle nativo Datadog (distinto de DD_APM_ENABLED):
+# false = ddtrace cargado pero sin enviar trazas (útil para debug local sin agente)
+#DD_TRACE_ENABLED=true
```
LucieFaire: could we please switch to english hehe :)

src/apm.py
```diff
@@ -0,0 +1,41 @@
...
+
+def trace(
+    operation_name: str,
+    service: str | None = None,
+    resource: str | None = None,
+) -> AbstractContextManager[Any]:
+    """Return a ddtrace span context manager, or a no-op when APM is disabled."""
+    if not ENABLED:
+        return nullcontext()
+    from ddtrace import tracer
```
LucieFaire: typecheck is failing becuase there is no tracer exposed on ddtrace.
we should either disable the check here
``` 
from ddtrace import tracer # type: ignore[attr-defined]
```
and remove the comment from L41

OR better export what we need: (no comment)
```
from ddtrace.trace import tracer
```
https://github.com/DataDog/dd-trace-py/blob/v4.10.5/ddtrace/trace/__init__.py

LucieFaire:
@claude review

src/clients/safe_client_gateway.py
```diff
@@ -42,15 +44,23 @@ def cgw_setup() -> tuple[str, str]:
...
-    try:
-        (url, token) = cgw_setup()
-        url = urljoin(url.rstrip("/") + "/", "v1/hooks/events")
-        payload: Dict[str, Any] = {"type": event.type, "chainId": str(event.chain_id)}
-        if event.service is not None:
-            payload["service"] = event.service
-        post(url, token, json=payload)
-    except Exception as error:
-        logger.error(error)
+    with apm.trace("cgw.hook_event", resource=event.type.value) as span:
```
LucieFaire: The `with apm.trace(...)` call, its `event.type.value` argument, and the `span.set_tag(...)` calls now run OUTSIDE the try/except, breaking hook_event's 'never raises to caller' contract when APM is enabled.

Example:
`DD_APM_ENABLED=true`, a SafeApp/Service save fires a `pre_save/post_save/post_delete receiver` (safe_apps/signals.py, chains/services.py) → `hook_event`. If `ddtrace's tracer.trace()` or `span.set_tag()` raises (uninitialized tracer, version mismatch, finished span), the exception escapes `hook_event`, propagates into the signal receiver, and aborts the model .save()/.delete() and its request — the exact fire-and-forget isolation the original full-body try/except guaranteed.

src/clients/safe_client_gateway.py
```diff
@@ -42,15 +44,23 @@ def cgw_setup() -> tuple[str, str]:
...
+        try:
+            (url, token) = cgw_setup()
+            url = urljoin(url.rstrip("/") + "/", "v1/hooks/events")
+            payload: Dict[str, Any] = {"type": event.type, "chainId": str(event.chain_id)}
+            if event.service is not None:
+                payload["service"] = event.service
+            post(url, token, json=payload)
+        except Exception as error:
+            if span is not None:
+                span.set_tag("error", True)
```
LucieFaire: `span.set_tag("error", True)` sets a meta tag named 'error' but does not set the span's reserved error state, so failed hooks appear successful in DD.

src/apm.py
```diff
@@ -0,0 +1,41 @@
+# SPDX-License-Identifier: FSL-1.1-MIT
+import logging
+import os
+from contextlib import AbstractContextManager, nullcontext
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+ENABLED = os.environ.get("DD_APM_ENABLED", "false").lower() == "true"
```
LucieFaire: DD_APM_ENABLED is parsed with a strict `== "true"`, so common truthy values (1, True, on, yes) silently disable APM; the repo already has a `_parse_bool` helper we can use.

diegopazosrego:
@claude review

diegopazosrego:
@claude review

src/clients/safe_client_gateway.py
```diff
@@ -43,14 +45,27 @@ def cgw_setup() -> tuple[str, str]:
...
+                    span.set_tag("cgw.service", event.service)
+            try:
+                (url, token) = cgw_setup()
+                url = urljoin(url.rstrip("/") + "/", "v1/hooks/events")
+                payload: Dict[str, Any] = {"type": event.type, "chainId": str(event.chain_id)}
+                if event.service is not None:
+                    payload["service"] = event.service
+                post(url, token, json=payload)
+            except Exception as error:
+                logger.exception(error)
```
LucieFaire, line 61: is it intentional that you switched from `.error` to `.exception` ? It can potentially add log volume.

uv.lock
```diff
@@ -15,11 +15,11 @@ safe-eth-py = false
 
 [[package]]
 name = "aiohappyeyeballs"
-version = "2.6.1"
+version = "2.6.2"
```
LucieFaire, line 18: hm, so many deps got upgraded, even though you only added one more dep. Are these all transient from ddtrace or you upgraded the uv lock?

LucieFaire (commented):
I am concerned about all the deps change in uv.lock but otherwise looks good

src/clients/safe_client_gateway.py
diegopazosrego: Intentional  logger.exception is the idiomatic Python choice inside except blocks: it logs at ERROR level (same number of events) but includes the full traceback automatically. Given this is a CGW hook failure path, having the stack trace in production logs is valuable for debugging. But we could switch to `.error` . Wdyt is the best way @LucieFaire ?

uv.lock
diegopazosrego: These are all transitive dependencies from ddtrace==4.10.5. ddtrace depends on aiohttp (async HTTP client used internally), which pulls in packages like aiohappyeyeballs. No manual upgrades were run  this is the output of uv lock after adding the single direct dependency. The lock file pins all transitive deps to ensure reproducibility.

src/clients/safe_client_gateway.py
LucieFaire: im all for having more data, lets keep :)

LucieFaire (commented):
🚀

LucieFaire (approved):

PR #1638, opened 2026-07-22
ci: relabel released PRs to 'autorelease: tagged' (locked-PR workaround)

diegopazosrego:
release-please publishes the Release but then fails to comment on the merged Release PR because the repo locks PRs on merge; it aborts before swapping the 'autorelease: pending' label to 'tagged', which blocks the next release (422 already_exists). Force the swap in a contained step (label edits are allowed on locked PRs). Idempotent; no org lock-policy change needed.

### DIFF
--- a/.github/workflows/_release.yml
+++ b/.github/workflows/_release.yml
@@ -81,3 +81,33 @@ jobs:
           set -euo pipefail
           PR_NUMBER=$(echo "$PR_JSON" | jq -r '.number')
           gh pr merge --auto --squash "$PR_NUMBER" --repo "$GITHUB_REPOSITORY"
+
+      # Safety-net for the recurring "stuck Release PR" problem.
+      # After publishing the Release, release-please tries to COMMENT on the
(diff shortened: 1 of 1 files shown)


LucieFaire (dismissed):
one qs: it is expected that it marks every pending PR as tagged? Maybe we should only mark the specific PR that was just merged?

diegopazosrego:
> one qs: it is expected that it marks every pending PR as tagged? Maybe we should only mark the specific PR that was just merged?

Good point — tightened it. It now relabels only the PR that was just released, looked up via the release commit SHA (steps.release.outputs.sha) → GET /commits/{sha}/pulls, instead of scanning all pending PRs

For context: the previous version already filtered to --state merged, so open/unreleased Release PRs were never touched — but scoping to the exact released PR is clearer and avoids any surprise if a stale autorelease: pending label ever lingered on an unrelated merged PR. Pushed as a follow-up commit.

PR #1681, opened 2026-09-10
ci: self-heal stuck Release PR labels regardless of release-please outcome

diegopazosrego:
## Problem

This repo locks PRs on merge (org-wide policy). After publishing a Release, release-please tries to **comment** on the just-merged Release PR, the comment fails (`issue is locked`), and the action aborts **before** flipping that PR's label `autorelease: pending` → `autorelease: tagged`. Left pending, the next run retries creating the already-published release and the whole `release-please` step **fails**.

The existing "Mark the released PR as tagged" safety-net was gated on `steps.release.outputs.release_created == 'true'` and used `steps.release.outputs.sha`. When release-please throws it sets no outputs, and a failed step skips later steps by default — so the safety-net was **dead code precisely in the scenario it was written for**. Both the Aug 27 and 2026-09-10 runs show it skipped, and the pipeline stayed red until the label was flipped by hand.

## Fix

Replace it with a self-healing step that runs `if: always()`, independent of release-please's outcome. It lists every **merged** Release PR still labeled `autorelease: pending` and, for each whose release tag **already exists**, flips it to `autorelease: tagged`. Genuinely unreleased PRs are left untouched. Editing labels is allowed on a locked PR (only commenting is blocked); the step is idempotent.

## Effect

A stuck Release PR label can no longer block releases: even when release-please fails on it, the heal step still runs and clears it, so the next run recovers on its own — no manual relabel needed.

### DIFF
--- a/.github/workflows/_release.yml
+++ b/.github/workflows/_release.yml
@@ -95,7 +95,7 @@ jobs:
       # 'tagged'. Editing labels is allowed on a locked PR (only commenting is
       # blocked). Idempotent; genuinely unreleased PRs are left untouched.
       - name: Heal stuck Release PR labels (locked-PR workaround)
-        if: ${{ always() && steps.app-token.outcome == 'success' }}
+        if: ${{ !cancelled() && steps.app-token.outcome == 'success' }}
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
         run: |
--- a/.github/workflows/_release.yml
+++ b/.github/workflows/_release.yml
@@ -82,34 +82,34 @@ jobs:
           PR_NUMBER=$(echo "$PR_JSON" | jq -r '.number')
           gh pr merge --auto --squash "$PR_NUMBER" --repo "$GITHUB_REPOSITORY"
 
-      # Safety-net for the recurring "stuck Release PR" problem.
-      # After publishing the Release, release-please tries to COMMENT on the
-      # just-merged Release PR — but this repo locks PRs on merge (org-wide
-      # policy), so the comment fails ("issue is locked") and the action aborts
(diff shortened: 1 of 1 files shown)


.github/workflows/_release.yml
```diff
@@ -82,34 +82,34 @@ jobs:
...
+      # 'autorelease: pending' -> 'autorelease: tagged'. Left pending, the NEXT run
+      # retries creating the already-published release and the whole job FAILS —
+      # which is exactly why a fix gated on release success is useless: it never
+      # runs (the failure happens first). So this step runs ALWAYS, independent of
+      # release-please's outcome: it finds every MERGED Release PR still on
+      # 'autorelease: pending' whose release tag ALREADY EXISTS and flips it to
+      # 'tagged'. Editing labels is allowed on a locked PR (only commenting is
+      # blocked). Idempotent; genuinely unreleased PRs are left untouched.
+      - name: Heal stuck Release PR labels (locked-PR workaround)
+        if: ${{ always() && steps.app-token.outcome == 'success' }}
```
LucieFaire: interesting, my Claude suggests:
```
${{ !cancelled() && steps.app-token.outcome == 'success' }}
```

LucieFaire (dismissed):

.github/workflows/_release.yml
diegopazosrego: Good call applied 👍

diegopazosrego: Fixed https://github.com/safe-global/safe-config-service/pull/1681/changes/eb0a7e7d657bdc227548432273bcdd38b5cc6244

LucieFaire (approved):
🚀

